Issue #4, opened 2017-01-06
Implement Two-Stage Methods


ChrisRackauckas:
From the video @finmod shared: http://www.birs.ca/events/2013/5-day-workshops/13w5151/videos/watch/201311140905-Campbell.html

Two-Stage methods just need to do a local linear regression on the data from the solutions. It should be a good beginner's issue. Note that [LossFunctions.jl](https://github.com/JuliaML/LossFunctions.jl) now has [weighted losses](https://github.com/JuliaML/LossFunctions.jl/issues/58) which should make the implementation simpler. This is a good beginner issue which only requires knowledge of using DifferentialEquations, with hints given by looking at the DiffEqParamEstim.jl source.

shivin9:
Can you give some references regarding this algorithm?

ChrisRackauckas:
Sorry I missed this request. These are the ones from the video.

https://arxiv.org/abs/0710.4190
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2631937/

ChrisRackauckas:
Honestly the references aren't as clear as Campbell's video.

ChrisRackauckas:
There is a PR in the works: https://github.com/JuliaDiffEq/DiffEqParamEstim.jl/pull/6

One interesting thing about this algorithm is that it just requires smoothing the data and estimating derivatives. The fact that it uses Loess is not central to the design. It seems like any other smoothed interpolation would work. It would be interesting to have other implementations which make use of that.

finmod:
I suggest focusing on #5 and how the build_loss objective replicates Equations 5 and 6 in the Robust and efficient parameter estimation paper. Of the 10 examples in AMIGO2, only 2 examples are simple or simplistic LSQ estimations. All others are likelihood estimations involving a metric and a covariance. Although it is equivalent to a nonlinear least square optimization problem, it is much more sophisticated than an OLS. The main focus of parameter estimation based on measure is how sophisticated equation 5 can be: uniform, linear, quadratic, Hellinger and many others. The NLS optimization problem 6 is always the same. All of this work is DiffEq model fitting to data and this is the only game in town. Curve fitting is a waste of time and resources.

ChrisRackauckas:
>I suggest focusing on #5 and how the build_loss objective replicates Equations 5 and 6 in the Robust and efficient parameter estimation paper

What paper are you referencing here?

[...]

PR #6, opened 2017-03-07
[WIP]: Two-stage method

Ayush-iitkgp:
This is an attempt to implement two-stage method #4 and reorganize the source code.
@ChrisRackauckas for the example:
```
using DifferentialEquations
using Optim
tpoints = [0.0,0.5,1.0]
data  = [1,exp(1),exp(2)]
pf_func = function (t,u,p)
    p*u
 end
pf = ParameterizedFunction(pf_func,[2])
u0 = [1.0]
tspan = (0.0,1.0)
prob = ODEProblem(pf,u0,tspan)
result = optimize(two_stage_method(prob,tpoints,data), -20.0, 20.0)
```

_**result.minimizer[1] should have been equal to 2 but it instead outputs -1.116706e+00**_

Any hints where I am going wrong?

Update - Two-stage method is not expected to converge to the actual parameter value and -1.11 is close to actual value 2. Looks like the implementation is correct.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,3 +5,4 @@ LossFunctions 0.0.2
 RecursiveArrayTools 0.0.2
 ForwardDiff
 Calculus
+Optim
--- a/src/DiffEqParamEstim.jl
+++ b/src/DiffEqParamEstim.jl
@@ -1,5 +1,6 @@
 module DiffEqParamEstim
 using DiffEqBase, LsqFit, LossFunctions, RecursiveArrayTools, ForwardDiff, Calculus
+
 include("two_stage_method.jl")
 
   ### LsqFit Method
--- a/src/DiffEqParamEstim.jl
+++ b/src/DiffEqParamEstim.jl
@@ -1,109 +1,11 @@
 module DiffEqParamEstim
 using DiffEqBase, LsqFit, LossFunctions, RecursiveArrayTools, ForwardDiff, Calculus
 
+
+include("lm_fit.jl")
+include("build_loss_objective.jl")
+include("build_lsoptim_objective.jl")
 include("two_stage_method.jl")
 
-  ### LsqFit Method
-  function lm_fit(prob::DEProblem,t,data,p0,alg;kwargs...)
-    f = prob.f
-
-    model = function (t,p)
-      for i in eachindex(f.params)
-        setfield!(f,f.params[i],p[i])
-      end
-      sol = solve(prob,alg;saveat=t,save_timeseries=false,dense=false,kwargs...)
-      y = vecvec_to_mat(sol.u)
-      vec(y)
-    end
-    curve_fit(model,t,vec(data),p0;kwargs...)
-  end
-
-  immutable DiffEqObjective{F,F2} <: Function
-    cost_function::F
-    cost_function2::F2
-  end
-  (f::DiffEqObjective)(x) = f.cost_function(x)
-  (f::DiffEqObjective)(x,y) = f.cost_function2(x,y)
-
-
-  function build_loss_objective(prob::DEProblem,t,data,alg;
-                                loss_func = L2DistLoss,
-                                mpg_autodiff = false,
-                                verbose = false,
-                                verbose_steps = 100,
-                                kwargs...)
-    f = prob.f
-    cost_function = function (p)
-      f = (t,u,du) -> prob.f(t,u,p,du)
-      uEltype = eltype(p)
-      u0 = [uEltype(prob.u0[i]) for i in 1:length(prob.u0)]
-      tspan = (uEltype(prob.tspan[1]),uEltype(prob.tspan[2]))
-      temp_prob = ODEProblem(f,u0,tspan)
-      if alg == nothing
-        sol = solve(temp_prob;saveat=t,save_timeseries=false,dense=false,kwargs...)
-      else
-        sol = solve(temp_prob,alg;saveat=t,save_timeseries=false,dense=false,kwargs...)
-      end
-      fill_length = length(t)-length(sol)
-      for i in 1:fill_length
-        push!(sol.u,fill(NaN,size(prob.u0)))
-        #push!(sol.u,zeros(prob.u0))
-      end
-      y = vec(vecvec_to_mat(sol.u))
-      norm(value(loss_func(),vec(data),vec(y)))
-    end
-
-    if mpg_autodiff
-      gcfg = ForwardDiff.GradientConfig(zeros(length(f.syms)))
-      g! = (x, out) -> ForwardDiff.gradient!(out, cost_function, x, gcfg)
-    else
-      g! = (x, out) -> Calculus.finite_difference!(cost_function,x,out,:central)
-    end
-
-    if verbose
-      count = 0 # keep track of # function evaluations
-    end
-
-    cost_function2 = function (p,grad)
-      if length(grad)>0
-        g!(p,grad)
-      end
-      
-      if verbose
-        count::Int += 1
-        if mod(count,verbose_steps) == 0
-          println("f_$count($p)")
-        end
-      end
-
-      cost_function(p)
-    end
-    DiffEqObjective(cost_function,cost_function2)
-  end
-
-  function build_lsoptim_objective(prob::DEProblem,t,data,alg;kwargs...)
-    internal_data = vec(data)
-    cost_function = function (p,out)
-      f = (t,u,du) -> prob.f(t,u,p,du)
-      uEltype = eltype(p)
-      u0 = [uEltype(prob.u0[i]) for i in 1:length(prob.u0)]
-      tspan = (uEltype(prob.tspan[1]),uEltype(prob.tspan[2]))
-      temp_prob = ODEProblem(f,u0,tspan)
-      if alg == nothing
-        sol = solve(temp_prob;saveat=t,save_timeseries=false,dense=false,kwargs...)
-      else
-        sol = solve(temp_prob,alg;saveat=t,save_timeseries=false,dense=false,kwargs...)
-      end
-      fill_length = length(t)-length(sol)
-      for i in 1:fill_length
-        push!(sol.u,fill(NaN,size(prob.u0)))
-        #push!(sol.u,zeros(prob.u0))
-      end
-      y = vec(vecvec_to_mat(sol.u))
-      out .= y.-internal_data
-    end
-  end
-
-  export lm_fit, build_loss_objective, build_lsoptim_objective, DiffEqObjective
 
 end # module
--- a/src/DiffEqParamEstim.jl
+++ b/src/DiffEqParamEstim.jl
@@ -1,5 +1,5 @@
 module DiffEqParamEstim
-using DiffEqBase, LsqFit, LossFunctions, RecursiveArrayTools, ForwardDiff, Calculus
+using DiffEqBase, LsqFit, LossFunctions, RecursiveArrayTools, ForwardDiff, Calculus, Optim
 
 
 include("lm_fit.jl")
--- a/src/DiffEqParamEstim.jl
+++ b/src/DiffEqParamEstim.jl
@@ -1,9 +1,6 @@
 module DiffEqParamEstim
-
-include("two_stage_method.jl")
-
 using DiffEqBase, LsqFit, LossFunctions, RecursiveArrayTools, ForwardDiff, Calculus
-
+include("two_stage_method.jl")
 
   ### LsqFit Method
   function lm_fit(prob::DEProblem,t,data,p0,alg;kwargs...)
--- a/src/DiffEqParamEstim.jl
+++ b/src/DiffEqParamEstim.jl
@@ -1,7 +1,7 @@
 __precompile__()
 
 module DiffEqParamEstim
-using DiffEqBase, LsqFit, LossFunctions, RecursiveArrayTools, ForwardDiff, Calculus, Optim
+using DiffEqBase, LsqFit, LossFunctions, RecursiveArrayTools, ForwardDiff, Calculus
 
 
 include("lm_fit.jl")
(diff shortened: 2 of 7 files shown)


ChrisRackauckas:
I think it would be good to find a way to abstract it a bit to not have to require Optim.jl, and also let MathProgBase optimizers be used.

Ayush-iitkgp:
Sure, would incorporate your suggestion. My first concern is to get it working for the toy example given above. Your help would be greatly appreciated.

Ayush-iitkgp:
**Some observations about this algorithm:**
1. It is fast w.r.t build_loss_objective method largely because it does not require us to solve the differential equation but at the same time, we need to solve the optimization problem for n times (where n is the size of the input data).
2. The estimates of the parameter are approximate.The recommended strategy is to use the two-stage method to get to the neighborhood of the actual parameter value and use it as the starting point of the build_loss_objective to converge to the actual parameter value.

@ChrisRackauckas I think we would need Optim as the dependency to solve the optimization problem in the fist step. Is there any other way around?

ChrisRackauckas:
What paper are you working from?

Ayush-iitkgp:
I have been using David Campbell's [talk](http://www.birs.ca/events/2013/5-day-workshops/13w5151/videos/watch/201311140905-Campbell.html) for the theory. The implementation is inspired by your work in build_loss_objective.

ChrisRackauckas:
Are you sure t0 isn't just fixed as the first timepoint? It looks like you loop through `i`, moving up t0 each time. The weights should take care of moving local time forward. 

One thing to note for later is that

```
ff = (t,u) -> prob.f(t,u,p)
```

is going to error for some definitions of the problem. It can either be `(t,u)` or the inplace `(t.u,du)`. You can handle that later though.

ChrisRackauckas:
Are there specialized optimization algorithms for local linear regression? I'd take a look at the internals of Loess.jl and see if you could use something from those kinds of libraries for optimizing that part. But yes, leave it as Optim until there's something working since it should work.

Ayush-iitkgp:
The first stage returns the value of beta0(t0) and beta1(t0) that is we get the estimates of x(t) and it's derivative at t=t0. 
But for the second stage, **we would need the estimates of x(t) and it's derivative for all the time points**.
Thus, the need to iterate first-stage for all time points. Am I thinking in the right direction?

> One thing to note for later is that
> `ff = (t,u) -> prob.f(t,u,p)`
> is going to error for some definitions of the problem. It can either be (t,u) or the inplace (t.u,du). You can handle that later though.

I have actually been wondering over how to solve this issue. We need the functional form of f(t,u) as given in the talk. Any idea on how to convert a function definition of the from f(t,u,du) to f(t,u)? OR the functional form of f(t,u,du) should work as well (in such case we would call ff as **ff(tpoints[i], b0[i], b1[i])**)

ChrisRackauckas:
>OR the functional form of f(t,u,du) should work as well 

Yes, `du=f(t,u)` is what it's like. Instead, you just have to pre-allocate and get the solution out from `f` which mutates. It's the same thing. There is a parameter on the ODEProblem for which function form it will be.

ChrisRackauckas:
>The first stage returns the value of beta0(t0) and beta1(t0) that is we get the estimates of x(t) and it's derivative at t=t0.
But for the second stage, we would need the estimates of x(t) and it's derivative for all the time points.
Thus, the need to iterate first-stage for all time points. Am I thinking in the right direction?

I see. Yes, it seems you are in the right direction. I "see" the whole algorithm now. It's probably a good idea to use Loess.jl here. If you want to finish up what you have here first, it's fine, but it seems Loess does have a specific optimization solution which can be faster in general, and throwing to the right package here could make things easier. The relevant field in the `LoessModel` is `bs`.

I would first check the output of the b's that your getting on really fine data. For `p=1`, you should have `b[0] = b[1]`. That's a good test case. If that's not what you're getting, then the b's are your problem. Make sure this is good before moving on. 

If this is good, then the problem is narrowed down to `cost_function2`.  At first glance it looks fine.

```julia
err = err + (ff(tpoints[i],b0[i]) - b1[i])^2
```

Note that Julia can optimize a tiny bit more if you use `err += (ff(tpoints[i],b0[i]) - b1[i])^2`

Ayush-iitkgp:
@ChrisRackauckas for the problem above for p = 1
I got 
```
julia> b0
3-element Array{Any,1}:
 0.905951
 1.80017 
 2.62445 

julia> b1
3-element Array{Any,1}:
 -1.77786
 -1.71832
 -1.65899

```

It looks like it's a bug but I think since the values are estimated values, it doesn't need to be b0=b1(since this method also says that it doesn't guarantee the exact value but somewhere in the covex region close to the exact value). What do you think?

ChrisRackauckas:
You shouldn't get negative b1's. Those are the local derivative estimates. They should always be positive.

I would suggest just dropping the Optim route here and using [Loess.jl](https://github.com/JuliaStats/Loess.jl). Let's take a step back a little bit. A bigger dataset which isn't exact but closer is likely easier to work with.

```julia
p = 1
tpoints = sort(rand(100))
data = exp(p*tpoints) + 0.1 * rand(100)
```

figure_1

Double check that Loess works:

```julia
model = loess(tpoints,data,degree=1)
us = collect(minimum(tpoints):0.1:maximum(tpoints))
vs = predict(model, us)
plot(us,vs)
```

figure_1-1

That produces a decent Loess prediction. This package gives the bs:

```julia
julia> model.bs
33×2 Array{Float64,2}:
 1.03091   1.2604
 0.821584  1.76706
 0.966791  1.45168
 1.03399   1.24084
 1.025     1.28773
 0.695939  1.98014
 1.03454   1.23721
 0.637211  2.07029
 0.649594  2.05219
 0.790347  1.82221
 ⋮
 0.770521  1.855
 0.571632  2.15629
 1.03372   1.24262
 1.02925   1.26919
 0.805961  1.79493
 0.713359  1.95179
 0.582638  2.14254
 1.03229   1.25201
 0.602776  2.11709
 0.606515  2.11226
```

I was wrong about `b0=b1` since `b0` is the intercept, not the constant. Sorry about that! Instead, what should happen is you need to shift the `b0` to get the estimate for `X(t)`. Since it's a line, it's just b0+b1*tpoint for each (tpoint,b0,b1)-tuple. That is what should be compared to `X'(t)`, which means that

```julia
err = err + (ff(tpoints[i],b0[i]) - b1[i])^2
```

this line is wrong. Instead, you need to use the X(t)-estimate that corresponds with the b1 == X'(t) estimate.

But:

1. I am not sure how to get the center timepoint out of the Loess model.
2. I am not sure why Loess.jl only gives 33 points, no matter how many points in the array I give.

Ayush-iitkgp:
I just went through the source code of Loess.jl, as far I understand, Loess takes a local interval and does a regression on it to find the parameters for that partiulat interval (interestingly it looks like they divide the overall intervall into 33 sub-intervals everytime). It is not very clear if they are doing the weighted linear regression. I would prefer doing something that I completely understand than trying a blackbox method that I have no idea of.

ChrisRackauckas:
>Loess takes a local interval and does a regression on it to find the parameters for that particular interval (interestingly it looks like they divide the overall intervall into 33 sub-intervals everytime). It is not very clear if they are doing the weighted linear regression.

Loess is a weighted linear regression. It's the common name for the algorithm. I guess they use a different weight function though. You'll even see on Campbell's next slide that he mentions you can just pull directly from an existing library for that step.

>Also, the weight function is not exponential in this algorithm.

This probably doesn't matter too much. As long as it's weighting for local windows. You may want to generalize this in your own algorithm. Likely the results aren't too dependent on the weight function.

>I am sorry, I don't understand why can't b1's be negative?

They are the local derivative estimates. In some window, the estimate is a local regression b0 + t*b1. All of your derivatives should be positive for that input.

src/two_stage_method.jl
```diff
@@ -0,0 +1,99 @@
...
+        T2 = construct_t2(tpoints[i],tpoints)
+        W = construct_w(tpoints[i],tpoints,h,kernel_function)
+        estimated_solution[i,:] = e1'*inv(T1'*W*T1)*T1'*W*data
+        estimated_derivative[i,:] = e2'*inv(T2'*W*T2)T2'*W*data
+    end
+
+    
+    # Step - 2
+    cost_function = function (p)
+        ff = (t,u,du) -> prob.f(t,u,p,du)
```
Ayush-iitkgp: @ChrisRackauckas help needed here?

ChrisRackauckas: that's good

src/two_stage_method.jl
```diff
@@ -0,0 +1,99 @@
...
+        #err = 0
+        sol = []
```
ChrisRackauckas: Type this: `sol = eltype(prob.u0)[]`

src/two_stage_method.jl
```diff
@@ -0,0 +1,99 @@
...
+        for i in 1:n
+            push!(sol,ff(tpoints[i],estimated_solution[i,:],estimated_derivative[i,:]))
```
ChrisRackauckas: The derivative estimates using X(t) and f:

```julia
du = similar(prob.u0)
for i in 1:n
  ff(tpoints[i],estimated_solution[i,:],du)
  push!(sol,du)
end
```

ChrisRackauckas: `estimated_solution[i,:]` Julia is column-major so this is really bad. Get it working, then go back and fix this.

ChrisRackauckas: `estimated_solution[i,:]` this also might only work on vector equations?

REQUIRE
```diff
@@ -5,3 +5,4 @@ LossFunctions 0.0.2
 RecursiveArrayTools 0.0.2
 ForwardDiff
 Calculus
+Optim
```
ChrisRackauckas: Is this still needed?

src/DiffEqParamEstim.jl
```diff
@@ -1,111 +1,14 @@
 __precompile__()
 
 module DiffEqParamEstim
+using DiffEqBase, LsqFit, LossFunctions, RecursiveArrayTools, ForwardDiff, Calculus, Optim
```
ChrisRackauckas: Drop Optim?

src/two_stage_method.jl
```diff
@@ -0,0 +1,98 @@
...
+        estimated_derivative[i,:] = e2'*inv(T2'*W*T2)T2'*W*data
+    end
+
+    
+    # Step - 2
+    cost_function = function (p)
+        ff = (t,u,du) -> prob.f(t,u,p,du)
+        sol = eltype(prob.u0)[]
+        for i in 1:n
+            push!(sol,ff(tpoints[i],estimated_solution[i,:]))
```
ChrisRackauckas: You dropped the `du` part.

src/two_stage_method.jl
Ayush-iitkgp: I am getting the following error
`ERROR: DimensionMismatch("arrays could not be broadcast to a common size")
`It's becuase ff is returning only the 1-element instead of 2-elements array whereas estimated_derivative is 2D array

ChrisRackauckas: Are you sure that's the issue? Print some stuff out.

src/two_stage_method.jl
```diff
@@ -0,0 +1,98 @@
+export two_stage_method
+
+# Step - 1
+function decide_kernel(kernel)
```
ChrisRackauckas: You should use symbols for this instead of strings.

src/two_stage_method.jl
```diff
@@ -0,0 +1,98 @@
...
+        for i in 1:n
+            push!(sol,ff(tpoints[i],estimated_solution[i,:],estimated_derivative[i,:]))
+        end
+        norm(value(loss_func(),vec(sol),vec(estimated_derivative)))
+    end
+    return cost_function
+end
+
+
+# cost_function = two_stage_method(prob,t,data)
```
ChrisRackauckas: add your test case to the tests file.

Ayush-iitkgp:
@ChrisRackauckas is it ready to get merged or some more modifications needed?

ChrisRackauckas:
There are no tests.

src/build_loss_objective.jl
```diff
@@ -11,24 +11,25 @@ end
...
-  fill_length = length(t)-length(sol)
-  for i in 1:fill_length
-    push!(sol.u,fill(NaN,size(prob.u0)))
-    #push!(sol.u,zeros(prob.u0))
-  end
-  y = vec(vecvec_to_mat(sol.u))
-  norm(value(loss_func(),vec(data),vec(y)))
+    f = (t,u,du) -> prob.f(t,u,p,du)
+    uEltype = eltype(p)
+    u0 = [uEltype(prob.u0[i]) for i in 1:length(prob.u0)]
```
ChrisRackauckas: you get bonus points for fixing my bad spacing.

test/runtests.jl
```diff
@@ -138,3 +138,9 @@ res = LeastSquaresOptim.optimize!(LeastSquaresOptim.LeastSquaresProblem(
                 ftol=1e-14,xtol=1e-15,iterations=100,grtol=1e-14)
 
 @test_approx_eq_eps res.minimizer [1.5;1.0;3.0;1.0] 3e-1
+
+
+println("Use Two Stage Method to fit the parameter")
+cost_function = two_stage_method(prob,t,data)
+result = Optim.optimize(cost_function.cost_function, 0.0, 10.0)
```
ChrisRackauckas: `Optim.optimize(cost_function, 0.0, 10.0)`. No need to pass the argument

src/two_stage_method.jl
```diff
@@ -1,6 +1,15 @@
-export two_stage_method
+export TwoStageCost, two_stage_method
+
+
+immutable TwoStageCost{F,D}
```
ChrisRackauckas: `immutable TwoStageCost{F,D} <: Function` will make it slightly easier to use.

ChrisRackauckas:
Thanks!

PR #7, opened 2017-03-19
WIP: Two stage method

shivin9:
An attempt to implement the two-stage method without using Optim.jl. Still it has some dimentionality issues though.
I am using curve_fit to find the optimal parameters beta_0 and beta_1.
@ChrisRackauckas do you think this can work?


ChrisRackauckas:
See https://github.com/JuliaDiffEq/DiffEqParamEstim.jl/issues/4#issuecomment-287623206

ChrisRackauckas:
Closing this because an implementation was just merged. However, there's still some to be done along these lines, like https://github.com/JuliaDiffEq/DiffEqParamEstim.jl/issues/10 .

PR #15, opened 2017-05-03
v0.6 test deps

ChrisRackauckas:
This fixes the depwarns in the v0.6 tests. Requires v0.6, so it should only be merged after v0.6 is required.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.6-
 LsqFit 0.0.2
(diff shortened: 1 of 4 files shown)


ScottPJones:
This needs the .travis.yml and appveyor.yml files updated to not try to test on v0.5.x any longer, since the minimum requirement for Julia has been set to v0.6

ChrisRackauckas:
Yeah, this won't merge for awhile.

PR #21, opened 2017-05-19
Update lorenz_true_test.jl

finmod:
The conclusions are substantially different with these changes in place. Many algorithms fail in the longer version.


ChrisRackauckas:
good catch. Which algorithms do well? Which don't? I'm going to try it out right now, but I'm curious.

ChrisRackauckas:
I just ran it and all of those NLopt algorithms still converge?

finmod:
For the short version, no problem. For the long version, not much is working. This seems to depend on the time step.

ChrisRackauckas:
>This seems to depend on the time step.

? But no timesteps are chosen?

finmod:
Use the following bounds, GloIniPar and LooserBounds for global algorithms and the others for local algorithms: 

Xiang2015Bounds = Tuple{Float64, Float64}[(9, 11), (20, 30), (2, 3)] # for local optimizations
LooserBounds = Tuple{Float64, Float64}[(0, 22), (0, 60), (0, 6)] # for global optimization
GloIniPar = [0.0, 0.5, 0.1] # for global optimizations
LocIniPar = [9.0, 20.0, 2.0] # for local optimization

Global optimization makes sense only if you do not know anything about initial values of parameters. In addition, global optimizers should not require starting values and BBO does not. 

What happen next is that for the short version, a few global algorithms succeed although some of the may only provide reasonable starting values for a local algo. For the long version, the global algos fail almost everywhere. Now with 500 obs or 3000 obs, I see no difference.

ChrisRackauckas:
>What happen next is that for the short version, a few global algorithms succeed although some of the may only provide reasonable starting values for a local algo.

That should be the case in harder problems.

>For the long version, the global algos fail almost everywhere. Now with 500 obs or 3000 obs, I see no difference.

:(. Even with the ODE solver tolerance reduced? If so, you've hit where the method meets its limit.

Good to know. So are there more robust algorithms designed to handle those cases? I think we're clear the the objective function works, but something more advanced than a nonlinear regression may be what's required in those cases. 

Do you have any methods to suggest? Are there any other packages you're testing this on to know how they fare? If so, what algorithms are they using to do this right?

The L2Loss is pretty simple:

https://github.com/JuliaDiffEq/DiffEqParamEstim.jl/blob/master/src/cost_functions.jl#L36

We should next check what happens when we throw some regularization terms in there. Etc. @Ayush-iitkgp is going a GSoC project on all of this, so if you have any suggestions for how exactly to tackle this problem let us know. In general, I don't handle chaotic equations so this is not a case I know very well, but once you get to these kinds of details you'll definitely need a method tailored to the type of equation.

finmod:
As you suggested, I will open several separate issues to continue improving this parameter estimation chapter. The curent version of the ipynb is available here: https://github.com/finmod/DiffEqParamEstim.jl/blob/master/test/lorenz%20true%20test.ipynb . Let me know if I need to do something else to give access to the ipynb.

ChrisRackauckas:
I checked the ProbInts

```julia
using DifferentialEquations, RecursiveArrayTools

g1 = @ode_def_nohes LorenzExample begin
  dx = σ*(y-x)
  dy = x*(ρ-z) - y
  dz = x*y - β*z
end σ=>10.0 ρ=>28.0 β=>2.6666

r0 = [0.1;0.0;0.0]
tspan = (0.0,30.0)
prob = ODEProblem(g1,r0,tspan)

cb = AdaptiveProbIntsUncertainty(5)
monte_prob = MonteCarloProblem(prob)
sim = solve(monte_prob,Tsit5(),num_monte=100,callback=cb,reltol=1e-9)

using Plots; plotly(); plot(sim)
```

newplot 5


```julia
summ = MonteCarloSummary(sim,linspace(0,30,1000))
plot(summ)
```

newplot 4

There doesn't seem to be any divergence in the numerical solution of the ODE, indicating that it's sufficiently exact enough to capture the results on that timespan. So it is just the optimization methods which are in need of help.

But you didn't change the optimizer settings at all. Most of them just hit max evaluations. I'll just change the true tests to have this setup, and see what I can get to do well.

finmod:
At this point, the diagnostic rests on visual inspection the residuals of the regression (not the mean square errors). A simple test for bias is to count the instances of positive and negative residuals. The estimator is unbiased if this count is zero. Next, see if there is heteroskedasticity (increasing or variable variance) of the residuals. I suspect that this is what is going on here since the time profile plotted above of the variables themselves is already heteroskedastic. We can see that the NLS estimator works over the [0,3] interval which is homoskedastic but then breaks down over [10,30] when the variance increases. 

This brings us back to what build_loss_objective is doing or not doing right now and what it should be doing. L2Loss is now the correct quadratic data misfit function of the earlier thread. It should be interchangeable for L1 or other robust data misfit distances.

Here, the NLS estimator is set to be exactly equal to the data misfit function i.e. NLS = L2Loss. If I understand well the NLS class is named the cost function in build_loss_objective. See my earlier submission https://github.com/finmod/DiffEqDocs.jl/blob/master/docs/src/analysis/Parameter%20estimation%20Lorenz%20bboptim.ipynb . The problem we face here is that the V matrix is just the identity matrix and fails. The weighted least square solution can be obtained by keeping the cost function as it is and transforming the data as indicated inside L2Loss or vice-versa keeping L2Loss as standard and performing the data transformation in the cost function.

At this point, the MLE estimator, also defined in the ipynb, is the most efficient estimator because the logarithmic transformation normalizes the heteroskedastic variance. One manner of defining a log-likelihood function in Julia can be found here:  https://github.com/kieranrcampbell/gpseudotime/blob/master/bgplvm.jl but I do not know if it is the most efficient and I could not identify an off-the-shelf function in the standard packages.

I suggest that the cost function should then take the name of the type of estimator: NLS or MLE and the data misfit function labelled by type: quadratic (L2Loss), linear, Huber and others.

ChrisRackauckas:
>I suspect that this is what is going on here since the time profile plotted above of the variables themselves is already heteroskedastic.

I suspect that for an entirely different reason: solutions at later timepoints are more sensitive to parameter choices than at earlier timepoints. You can verify this with local sensitivity analysis:

http://docs.juliadiffeq.org/latest/analysis/sensitivity.html#Local-Sensitivity-Analysis-1

A small change in the parameters can have a large change at `t=20`, even if the change at `t=1` is minimal. 

>L2Loss is now the correct quadratic data misfit function of the earlier thread. It should be interchangeable for L1 or other robust data misfit distances.

Yes, you can define any loss you want here. Or `CostVData` and pass in other loss functions. But without defining weights, these will not take into account heteroskedasticity. 

>Here, the NLS estimator is set to be exactly equal to the data misfit function i.e. NLS = L2Loss. If I understand well the NLS class is named the cost function in build_loss_objective. 

Yes.

>See my earlier submission https://github.com/finmod/DiffEqDocs.jl/blob/master/docs/src/analysis/Parameter%20estimation%20Lorenz%20bboptim.ipynb .

One thing to note in there is that the SDE solution is not the same as (1) BTW.

>The problem we face here is that the V matrix is just the identity matrix and fails. The weighted least square solution can be obtained by keeping the cost function as it is and transforming the data as indicated inside L2Loss or vice-versa keeping L2Loss as standard and performing the data transformation in the cost function.

The loss function should take in a vector of weights. Why are you using a weight matrix instead of a vector? Incorporating correlations of errors in time?

>At this point, the MLE estimator, also defined in the ipynb, is the most efficient estimator because the logarithmic transformation normalizes the heteroskedastic variance. One manner of defining a log-likelihood function in Julia can be found here: https://github.com/kieranrcampbell/gpseudotime/blob/master/bgplvm.jl but I do not know if it is the most efficient and I could not identify an off-the-shelf function in the standard packages.

@Ayush-iitkgp is going to handle this in his GSoC project.

finmod:
@ChrisRackauckas I have just discovered that the issue of precision of the solver for Lorenz system as T increases is well documented in this paper by Kehlet and Logg : https://www.simula.no/file/simulaacdc22pdf/download  . 

The direction to take for a solution is to use Galerkin-family solvers like the continuous and discontinuous Galerkin methods cG(q) and dG(q), formu-
lated by requiring that the residual R = ˙U −f(U, ·) be orthogonal to a suitable space
of test functions. In Julia, you already know of the extensive work undertaken by jInv.jl but only for PDEs so far. Can the family of solvers in JuliaDiffEq be extended to some of these for ODEs and SDEs?

ChrisRackauckas:
>The direction to take for a solution is to use Galerkin-family solvers like the continuous and discontinuous Galerkin methods cG(q) and dG(q), formu-
lated by requiring that the residual R = ˙U −f(U, ·) be orthogonal to a suitable space
of test functions. In Julia, you already know of the extensive work undertaken by jInv.jl but only for PDEs so far. Can the family of solvers in JuliaDiffEq be extended to some of these for ODEs and SDEs?

Yes, that's very similar to the direction we are taking in the BVP and neural net GSoC projects. Galerkin methods are another "parallel-in-time" method, where you can think of BVP collocation methods as just hat function methods in time. We're working on this stuff, but getting them to converge efficiently is difficult. However, the NeuralNetDiffEq and TensorFlowDiffEq stuff is interesting because it's "hot-startable", meaning that after training the net for one solution, you can use the same neural net parameters to train a different parameter set, effectively starting much closer to the desired solution in increasing convergence rates. We will have lots of work to do there though. That's just doing a standard collocation, but the NN methods directly give Galerkin methods if they minimize some basis instead.

But parallel-in-time is much more difficult than doing this in space (i.e. standard PDEs) because of how the  dependencies and error growth works (you only know start times!)

PR #23, opened 2017-06-06
Adding more Kernels for local regression in the 1st stage of TwoStageMethod

Ayush-iitkgp:
TwoStageMethod now supports all the kernels mentioned here: https://en.wikipedia.org/wiki/Kernel_(statistics)#Kernel_functions_in_common_use

### DIFF
--- a/src/DiffEqParamEstim.jl
+++ b/src/DiffEqParamEstim.jl
@@ -10,7 +10,7 @@ include("lm_fit.jl")
 include("build_loss_objective.jl")
 include("build_lsoptim_objective.jl")
 include("two_stage_method.jl")
-
+include("kernels.jl")
 export DECostFunction, CostVData, L2Loss
 
 end # module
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
This kernel function choice isn't type stable. You'll either want to put a function barrier in there or make the function choice due to a type. Or allow the user to directly pass the function as a default arg.

Ayush-iitkgp:
@ChrisRackauckas suggestion incorporated.

ChrisRackauckas:
Yeah that should be fully inferred. I'll merge when it's all green.

Ayush-iitkgp:
@ChrisRackauckas any reasons why it is taking so long for AppVeyor build to complete?

ChrisRackauckas:
I re-enabled a bunch of leastsquaresoptim tests. I'll merge this, then we should think about refactoring the tests somehow since it's getting unruly and some of it may be over-tested.

ChrisRackauckas:
Actually, the bigger thing is it looks like the AppVeyor build is bugged. I'll look into that later.

Issue #26, opened 2017-06-08
Add support for optimizing the cost function with BlackBoxOptim.jl


Ayush-iitkgp:
Add support for optimizing the cost function using global optimization framework [BlackBoxOptim.jl](https://github.com/robertfeldt/BlackBoxOptim.jl)

ChrisRackauckas:
This is already supported. It just needs tests and docs.

finmod:
Whereas everything in DiffEqParamEstim complies with Julia v0.6, BlackBoxOptim does not yet. @robertfeldt

ChrisRackauckas:
I don't think there's anything more to this on our end.

Issue #25, opened 2017-06-07
Support for Genetic Algorithm to optimize the cost function


Ayush-iitkgp:
[This paper](https://bmcsystbiol.biomedcentral.com/articles/10.1186/s12918-015-0219-2) suggests using heuristic techniques to find the global optima. These methods usually obtain an adequate solution in reasonable computation time, although at the price of no guarantee. The first step is to support Genetic Algorithms. Julia has the following GA libraries:
1. [Evolutionary.jl](https://github.com/wildart/Evolutionary.jl)
2. [GeneticAlgorithms.jl](https://github.com/WestleyArgentum/GeneticAlgorithms.jl)

ChrisRackauckas:
#28

Issue #12, opened 2017-04-07
Test with Evolutionary.jl


ChrisRackauckas:
https://github.com/wildart/Evolutionary.jl

This is a set of evolutionary and genetic algorithms. It would be good to make sure we're compatible.

Ayush-iitkgp:
#25 my bad I opened a similar issue. The new PR adds test for GA.

ChrisRackauckas:
#28

PR #28, opened 2017-06-08
Refracted test of ODEs and added tests for GA and BlackBoxOptim

Ayush-iitkgp:
Test for genetic algorithm from Evolutionary.jl (issue #12 and #25 ) and BlackBoxOptim (issue #26 )


Ayush-iitkgp:
Please note that we need to first checkout to master branch of both Evolutionary.jl and BlackBoxOptim.jl to use it with Julia 0.6+

ChrisRackauckas:
Then comment out those testsets for now. Probably should add a note that it's because master is needed. Tests should be on released versions only unless there's a really specific need.

Issue #29, opened 2017-06-08
Implement maximum likelihood function


Ayush-iitkgp:
The parameter estimation problem is basically to find the parameters in order to maximize the likelihood of the occurrence of the observed data. We assume that the samples are drawn from the Normally distributed population and each sample has its own variance. I would like to invite the concerned parties to express what there expectations are before I start writing the code. 
I plan to use [Robust and efficient parameter estimation in dynamic models of biological systems](https://bmcsystbiol.biomedcentral.com/articles/10.1186/s12918-015-0219-2) paper. Feel free to mention other sources as well.
Tagging few people I know @finmod, @ChrisRackauckas and @gabrielgellner others feel free to join in :)

Ayush-iitkgp:
Since each observation is drawn from independent normal distributions with different variances, we will need to ask the user to pass the variances of the populations as the argument to the API. Would be a right to pass the default argument as a vector of ones?

ChrisRackauckas:
I don't see a reason to constrain it to normal distribution only. Actually, as part of a research project I want to start testing other distributions. One which could be very interesting is t-distributions: this would naturally make it more robust to outliers. I think you can set something up where you use a distribution from Distributions.jl to do this.

But that is secondary though. Just like CostVData vs L2Loss, we will want a specialized fast version for the normal distribution anyways since that's the most common. So I would say start yes get that one first, but we shouldn't throw out the idea of something more general later.

ChrisRackauckas:
>Would be a right to pass the default argument as a vector of ones?

If all of the datapoints are in the `1e-6` range then the default would be way off. I think a good estimate would be `var(data)*ones(N)`. I say that without any proof or anything to back it up other than it seems that's the best a priori estimate 👍

ChrisRackauckas:
Also, note that the variances for the errors should always grow in any non-oscillatory example. In the general case they can grow exponentially (chaotic systems), so no math will give a good upper bound here. My hunch here is that guessing a linear growth with the slope related to the local mean/ local var of the data would be reasonable, though a constant guess is probably better to start with.

finmod:
[...]

Issue #33, opened 2017-06-14
Add Tikhonov regularization term to cost function in build_loss_objective function


Ayush-iitkgp:
As mentioned in Robust and efficient parameter estimation paper. Regularization is used to surmount ill-posedness and ill-conditioning of the problem. `MahalanobisPenalty` from [PenaltyFunctions.jl](https://github.com/JuliaML/PenaltyFunctions.jl) already offers this functionality. The only thing that is needed to be done is show the use case with an example and add docs. I am looking for examples with regard to differential equations where this technique is useful.

ChrisRackauckas:
```julia
p = MahalanobisPenalty(C)
value(p, x)
```

That should evaluate it on `x`.

Ayush-iitkgp:
```
penalty = MahalanobisPenalty(W)
regularization_term = value(penalty,p) # p is the unknown parameter of DE 
regularized_cost_function = non_regularized_cost_function + regularization_term
```
I think the above code would work and return the cost_function to be optimized by different optimizers.

ChrisRackauckas:
You probably want a lambda in there to scale the regularization term?

Ayush-iitkgp:
Oh! yes. Any use case you have in mind? Looks easy to implement. I will make a PR as soon as we have an use case.

ChrisRackauckas:
Use case as in test case? AMIGO2 uses Fitzhugh-Nagumo for this. 

https://sites.google.com/site/amigo2toolbox/examples

For everything, our 3 simple tests should be:

1) Lotka-Volterra
2) Fitzhugh-Nagumo
3) Hogskin-Huxley

Those are classic models every systems biologist would know, and are relatively sane, making them good test cases. Then of course benchmark on larger systems, but those 3 are good small cases.

Ayush-iitkgp:
Should be closed?

ChrisRackauckas:
Yes.

PR #34, opened 2017-06-14
Implement maximum_likelihood function

Ayush-iitkgp:
This the first implementation of the likelihood function. The objective is to share the code and ask for reviews and improve upon it.
Issue #29


src/build_maximum_likelihood_objective.jl
```diff
@@ -0,0 +1,43 @@
...
+  maximum_likelihood_function = function (p)
+    if verbose_opt
+      count::Int += 1
+      if mod(count,verbose_steps) == 0
+        println("f_$count($p)")
+      end
+    end
+    tmp_prob = prob_generator(prob,p)
+    sol = solve(tmp_prob,alg;saveat=likelihood.t,save_everystep=false,dense=false,kwargs...)
+    likelihood(sol)
```
ChrisRackauckas: Why is this a different outer function? The likelihood is just a different cost function, so this should be unnecessary.

src/cost_functions.jl
```diff
@@ -42,3 +42,25 @@ end
...
+function (f::MaximumLikelihood)(sol::DESolution)
+  data = f.data
+  variance = f.variance
+  fill_length = length(f.t)-length(sol)
+  for i in 1:fill_length
+    push!(sol.u,fill(Inf,size(sol[1])))
+  end
+  prod = 1.0
+  @inbounds for i in 1:length(sol)
+    for j in 1:length(sol[i])
```
ChrisRackauckas: you can make double-loops in one line

src/cost_functions.jl
```diff
@@ -42,3 +42,25 @@ end
...
+end
+
+function (f::MaximumLikelihood)(sol::DESolution)
+  data = f.data
+  variance = f.variance
+  fill_length = length(f.t)-length(sol)
+  for i in 1:fill_length
+    push!(sol.u,fill(Inf,size(sol[1])))
+  end
+  prod = 1.0
```
ChrisRackauckas: make this match the type of the solution, i.e. `one(eltype(sol.u))`

src/cost_functions.jl
ChrisRackauckas: for i in 1:length(sol), j in 1:length(sol[i])

src/cost_functions.jl
```diff
@@ -42,3 +42,25 @@ end
...
+  @inbounds for i in 1:length(sol)
+    for j in 1:length(sol[i])
+      prod *= exp(((data[j,i] - sol[j,i])^2)/(-2*variance[j,i]))/(sqrt(2*π*variance[j,i]))
```
ChrisRackauckas: Does the compiler elide the 2*π or compute it each round? You might want to save that as a constant. Also, this line is a very good candidate for an `@fastmath` annotation to speed up the `exp` and `sqrt`. The extra error essentially acts as regularization anyways.

ChrisRackauckas:
Made some style points for the implementation, but other than those I think it's pretty good. Do you have tests on the basic Lotka-Volterra problem yet?

src/build_maximum_likelihood_objective.jl
Ayush-iitkgp: You mean to replace `likelihood` by `loss`?

ChrisRackauckas: The only difference between the two functions is you changed the name of the loss function. Won't the likelihood function work with the loss objective? There's no reason to duplicate code to change a few names.

ChrisRackauckas:
From what I can see, `variance` is not used as a parameter, and so it will not be optimized. Don't you want to treat those as non-constant hyperparameters to train along with the other parameters?

src/build_maximum_likelihood_objective.jl
Ayush-iitkgp: `loss` takes 3 or 2 arguments(data,t and optional loss_function) whereas likelihood takes 3 arguments (data, t, and variance of the population from which the data has been drawn).
Does that make sense to give it a different name?

ChrisRackauckas: Their constructors are different, but their calls are the same.

Ayush-iitkgp:
No, the user has to pass the `variance` as the argument to build_maximum_likelihood_objective https://github.com/csynbiosys/AMIGO2R2016b/blob/master/Examples/ParameterEstimation_PE/NFKB_PE.m#L141

ChrisRackauckas:
>No, the user has to pass the variance as the argument to build_maximum_likelihood_objective 

Then why don't you do that? You're having it passed to the likelihood construction instead. I agree that it needs to be passed in in order to be optimized as noted earlier. But you can just recognize that it's a likelihood function via some abstract type and then use that to append the variances as part of the parameters.

Ayush-iitkgp:
This is the reason why I named it `likelihood` instead of `loss`. Just to differentiate the definitions.
You want it to be explicitly passed to build_maximum_likelihhod_objective?

ChrisRackauckas:
Either way is fine. You can either check if it's a likelihood function and grab the initial variances, or add that they can be passed in. Either way, there's one spot that handles the variances, but the rest should be exactly the same, right?

Also, don't look at the AMIGO2 code. It's GPL licensed. You're breaking the law if any of your code is derivative of it. You can use their paper as a reference, but do not use their code as a reference.

ChrisRackauckas:
So why exactly does `build_maximum_likelihood_objective` have to be a different function than `build_loss_objective`? It does everything the same?

Ayush-iitkgp:
The most recent commit also addresses regularization issue #33

Ayush-iitkgp:
@ChrisRackauckas if you are satisfied, I will modify the tests accordingly.

Ayush-iitkgp:
The new API would be:
```
W = eye(sizeof(p))
penalty = MahalanobisPenalty(W)
lambda = 0
cost_function = build_loss_objective(prob,Tsit5(),CostVData(t,data),Regularization(lambda,penalty),maxiters=10000)
```
lambda = 0 implies no regularization

ChrisRackauckas:
`lambda = 0 implies no regularization`

Why not just have it as a default argument:

```julia
function build_loss_objective(prob::DEProblem,alg,loss,regularization = nothing
```

This way, if the user doesn't pass it, it becomes type information and you can compile away that branch. Then:

```julia
if regularization == nothing
  return loss(sol)
else 
  return loss(sol) + regularization(p)
end
```

will compute which of the two it should do when the arguments are passed in.

ChrisRackauckas:
Yeah, other than that style point, looks good.

src/build_loss_objective.jl
```diff
@@ -8,7 +8,7 @@ end
 (f::DiffEqObjective)(x) = f.cost_function(x)
 (f::DiffEqObjective)(x,y) = f.cost_function2(x,y)
 
-function build_loss_objective(prob::DEProblem,alg,loss,regularization;mpg_autodiff = false,
+function build_loss_objective(prob::DEProblem,alg,loss;regularization=nothing,mpg_autodiff = false,
```
ChrisRackauckas: default argument, not a keyword argument. `build_loss_objective(prob::DEProblem,alg,loss,regularization=nothing;`, see the difference with the `;`. Keyword arguments won't dispatch so that'll be somewhat of an issue (though actually not that big of a deal). But this will feel more natural, i.e. you usually leave it off but add it when you want.

Ayush-iitkgp:
@ChrisRackauckas the maximum_likelihood_cost_function is taking forever to converge with NLopt and very poor convergence for Genetic Algorithm (0.72). Would you also want to look into it?

ChrisRackauckas:
Forever in terms of time or number of iterations?

Ayush-iitkgp:
In terms of time.

ChrisRackauckas:
But the number of iterations are similar? So it's a code optimization problem?

Ayush-iitkgp:
I think it's the problem with the complexity of the likelihood function given that it is exponential. In fact, this is the reason why we construct log-likelihood at the first stage.

ChrisRackauckas:
But does it take the same number of iterations, just each iteration taking longer to compute?

Ayush-iitkgp:
Nothing much has changed by negation. The maximum_likelihood_cost_function is still not converging. Are we doing anything wrong? @ChrisRackauckas @finmod any views?

ChrisRackauckas:
Have you changed the `variance` around?

PR #37, opened 2017-07-21
Adding support for heauristic methods - PSO, Simulated annealing and NelderMead

Ayush-iitkgp:
Note: NealderMead method is only for multivariate optimization.


ChrisRackauckas:
A test failed on Mac?

PR #40, opened 2017-07-23
Support for Regularization

Ayush-iitkgp:
This implements support for [regularization](https://github.com/JuliaML/PenaltyFunctions.jl/blob/master/README.md#available-penalties) terms upported in [PenaltyFunctions.jl](https://github.com/JuliaML/PenaltyFunctions.jl/blob/master/README.md).


src/build_loss_objective.jl
```diff
@@ -23,12 +23,17 @@ function build_loss_objective(prob::DEProblem,alg,loss;mpg_autodiff = false,
       end
     end
     tmp_prob = prob_generator(prob,p)
-    if typeof(loss) <: Union{CostVData,L2Loss}
+    if typeof(loss) <: Union{CostVData,L2Loss,MaximumLikelihood}
```
ChrisRackauckas: MaximumLikelihood isn't in this PR.

src/cost_functions.jl
```diff
@@ -1,3 +1,15 @@
+export DECostFunction, CostVData, L2Loss, MaximumLikelihood, Regularization
```
ChrisRackauckas: MaximumLikelihood isn't in this PR.

ChrisRackauckas:
Are there tests you're adding to this?

Ayush-iitkgp:
I wonder what the test would be like?
I can add the documentation on how to use it.

Ayush-iitkgp:
Tests failing? Looks like travis setup has some problems.

ChrisRackauckas:
It needs PenaltyFunctions in the REQUIRE file.

Ayush-iitkgp:
Ready to merge?

ChrisRackauckas:
No tests on this? Should at least show that some common methods still converge to the correct parameters with this. Have you created any plots of the 1D cost surface to show that it indeed smooths it? I don't think that is possible to do as a unit test but it would be a good check to have around.

src/cost_functions.jl
```diff
@@ -1,3 +1,15 @@
+export DECostFunction, CostVData, L2Loss, Regularization
+
+type Regularization{L,P} <: DECostFunction
+  λ::L
+  penalty::P
+end
+
+function (f::Regularization)(p)
```
Ayush-iitkgp: MahalanobisPenalty works on arrays so if the parameter estimation problem is to find a single parameter, it spits out the error. Any workaround?

ChrisRackauckas: I don't think we need to work around it. If that penalty is used on a scalar it's just used incorrectly.

ChrisRackauckas:
Nice. I resolved the conflict added due to a printing change. Will merge when tests pass.

PR #47, opened 2017-08-08
Remove duplicate calculation of loss value

devmotion:
I guess the loss value of a solution is calculated twice at the moment if a regularization is specified.


ChrisRackauckas:
Oops haha

PR #50, opened 2017-08-16
Multiple shooting method

Ayush-iitkgp:
This is the more refracted and complete version of Multiple Shooting Method. 
We have to make the objective function and the constraints NLOpt compatible in order to get it working.


src/multiple_shooting_method.jl
```diff
@@ -0,0 +1,74 @@
...
+  constraints::C
+end
+
+my_problem_new_parameters = function (prob::ODEProblem,p,t1,t2,data)
+  f = (t,u,du) -> prob.f(t,u,p,du)
+  uEltype = eltype(p)
+  tspan = (uEltype(t1),uEltype(t2))
+  u0 = [uEltype(data[:,t1][i]) for i in 1:length(prob.u0)]
+  ODEProblem(prob.f,u0,(t1,t2))
+end
```
ChrisRackauckas: Only works for ODEs but this can get cleaned up later. Should open an issue on this.

src/multiple_shooting_method.jl
```diff
@@ -0,0 +1,74 @@
...
+    if typeof(loss) <: Union{CostVData,L2Loss}
+      sol = solve(tmp_prob,alg;saveat=loss.t,save_everystep=false,dense=false,kwargs...)
+    else
+      sol = solve(tmp_prob,alg;kwargs...)
+    end
+    multiple_shooting_cost +=loss(sol)
+    # Construct the constraints
+    if boundary_condition != nothing
+      for j in 1:length(sol(t1))
+        push!(constraints, boundary_condition[j]=sol(t1)[j])
```
ChrisRackauckas: doesn't NLopt give you a constraint vector to fill?

ChrisRackauckas:
Needs tests.

Ayush-iitkgp:
I have made the changes and added test but it's not working somehow. @ChrisRackauckas would you want to have a look at it?

src/multiple_shooting_method.jl
```diff
@@ -0,0 +1,108 @@
...
+                              prob_generator = problem_new_parameters,
+                              kwargs...)
+  if verbose_opt
+    count = 0 # keep track of # function evaluations
+  end
+  multiple_shooting_cost = 0.0
+  constraints = []
+  data = loss.data
+  boundary_condition = nothing
+  if timestamp == nothing
```
ChrisRackauckas: what's timestamp and why does it have to double the code?

src/multiple_shooting_method.jl
```diff
@@ -0,0 +1,108 @@
...
+      count::Int += 1
+      if mod(count,verbose_steps) == 0
+        println("f_$count($p)")
+      end
+    end
+    tmp_prob = my_problem_new_parameters(prob,p,t1,t2,data)
+    if typeof(loss) <: Union{CostVData,L2Loss}
+      sol = solve(tmp_prob,alg;saveat=loss.t,save_everystep=false,dense=false,kwargs...)
+    else
+      sol = solve(tmp_prob,alg;kwargs...)
```
ChrisRackauckas: why is the ODE being solved again in the constraints?

src/multiple_shooting_method.jl
Ayush-iitkgp: timestamp is an array which act as the boundary point of the intervals. If the user does not input the intervals, we default it to 10 intervals hence, the condition.

ChrisRackauckas: then why don't you just default `timestamp = linspace(prob.tspan[1],prob.tspan[end],10)`?

src/multiple_shooting_method.jl
Ayush-iitkgp: This is the only time when the ODE is being solved.

ChrisRackauckas: oh I see

ChrisRackauckas:
What exactly is the issue?

Ayush-iitkgp:
`multiple_shooting_objective_constraints = multiple_shooting_method(prob1,Tsit5(),L2Loss(t,data),maxiters=10000)`
**multiple_shooting_objective_constraints.multiple_shooting_cost = 0.0?**
Similarly, multiple_shooting_objective_constraints.constraints = Any[0]?

ChrisRackauckas:
it's not returning the functions?

Ayush-iitkgp:
Yes

test/tests_on_odes/multiple_shooting_test.jl
```diff
@@ -0,0 +1,14 @@
+using NLopt
+
+println("Use Multiple Shooting Method the parameter")
```
ChrisRackauckas: this test isn't enabled though?

ChrisRackauckas:
I'm not sure this is correct so I'm closing.

PR #52, opened 2017-08-24
Adding weighted loss functionality to L2Loss and CostVData

Ayush-iitkgp:
Adding weighted loss functionality to L2Loss and CostVData

### DIFF
--- a/src/cost_functions.jl
+++ b/src/cost_functions.jl
@@ -25,7 +25,7 @@ function (f::CostVData)(sol::DESolution,weight)
   if weight == nothing
     weight = ones(length(vec(f.data)))
   end
-  norm(value(f.loss_func(),vec(f.data),vec(sol)).*f.weight)
+  norm(value(f.loss_func(),vec(f.data),vec(sol)).*vec(weight))
 end
 
 function (f::CostVData)(sol::AbstractMonteCarloSolution)
@@ -39,7 +39,7 @@ struct L2Loss{T,D} <: DECostFunction
   data::D
 end
 
-function (f::L2Loss)(sol::DESolution)
+function (f::L2Loss)(sol::DESolution,weight)
   data = f.data
   fill_length = length(f.t)-length(sol)
   for i in 1:fill_length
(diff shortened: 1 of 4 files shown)


src/cost_functions.jl
```diff
@@ -36,16 +39,20 @@ struct L2Loss{T,D} <: DECostFunction
...
     push!(sol.u,fill(Inf,size(sol[1])))
   end
+  if weight == nothing
+    weight = ones(f.data)
+  end
   sumsq = 0.0
   @inbounds for i in 1:length(sol)
     for j in 1:length(sol[i])
       sumsq += (data[j,i] - sol[j,i])^2
+      sumsq *=weight[j,i]
```
ChrisRackauckas: this will apply the weight to all previous values

ChrisRackauckas:
I think weights should just be added as part of those loss functions, not to the objective function itself. The thing which shows that is the fact that this code directly sends it to the loss function and no where else.

Ayush-iitkgp:
The default value of weight would be the matrix of ones which have the same size as of the data. Hence, it can't be an optional argument to the constructor. Would it be a right idea to ask the default values to the user?

ChrisRackauckas:
Why not just default `nothing` and then handle that? I would make the no-weight case have less operations anyways.

ChrisRackauckas:
Wait, why can't it be a default arg? One of the previous args has to be the data, so you can use that.

ChrisRackauckas:
Needs a test.

Ayush-iitkgp:
Test added and passing on Mac.

test/tests_on_odes/weighted_loss_test.jl
```diff
@@ -0,0 +1,24 @@
...
+  tmp = rand()
+  weight[i,j] = tmp*tmp
+  d = Normal(0,tmp)
+  error[i,j] = rand(d)
+ end
+end
+
+weighted_data = original_solution_matrix_form + error
+
+weighted_cost_function = build_loss_objective(prob1,Tsit5(),CostVData(t,weighted_data',weight=weight'),maxiters=10000)
```
ChrisRackauckas: why the transposes?

PR #53, opened 2017-08-25
Integrating Distributions.jl to output loglikelihood cost function

Ayush-iitkgp:
This PR demonstrates how to integrate Distributions.jl to find log likelihood cost function by taking a case when all the observations independently follow Normal Distribution. 
This would act as the guideline to integrate other cases as the need arise.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -8,3 +8,4 @@ Calculus
 Optim
 Compat 0.17.0
 PenaltyFunctions
+Distributions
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
In order to calculate log likelhoods you need to fit against multiple points at a time. This won't be possible with just an ODE and would require a MonteCarloProblem input.

Ayush-iitkgp:
If I am getting it right, I guess only the data would vary over different simulations of MonteCarloProblem so data would be matrices stacked with each other(if the number of simulations=100 then we need 100 data matrices).

ChrisRackauckas:
Yup exactly. And it needs to be stochastic otherwise it'll all be the same value.

ChrisRackauckas:
Completed in another Pr.

PR #54, opened 2017-08-26
Integrating Stan with DiffEqParamEstim

Ayush-iitkgp:


ChrisRackauckas:
This would blow up our dependency list. I think this should be a PR to DiffEqBayes.jl instead since both will need Mamba.jl anyways.

Ayush-iitkgp:
Moved [here](https://github.com/JuliaDiffEq/DiffEqBayes.jl/pull/3)

PR #55, opened 2017-10-14
Fix autodiff setup

devmotion:
Cost function was missing in the constructor of a `GradientConfig`. Moreover, I was wondering whether it might be reasonable to allow other initial vectors (for the autodifferentiation setup; currently it always creates a `Vector{Float64}`) and also custom chunk sizes, maybe as keyword arguments?

And regarding differentiation: I'm not up to date since I'm quite busy at the moment - but didn't you want to get rid of Calculus dependencies? Just noticed that the finite differentiation alternative still uses it here.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,5 +6,4 @@ RecursiveArrayTools 0.10.0
 ForwardDiff
 Calculus
 Optim
-Compat 0.17.0
 PenaltyFunctions
--- a/src/DiffEqParamEstim.jl
+++ b/src/DiffEqParamEstim.jl
@@ -3,8 +3,6 @@ __precompile__()
 module DiffEqParamEstim
 using DiffEqBase, LsqFit, LossFunctions, PenaltyFunctions, RecursiveArrayTools, ForwardDiff, Calculus
 
-using Compat
-
 include("cost_functions.jl")
 include("lm_fit.jl")
 include("build_loss_objective.jl")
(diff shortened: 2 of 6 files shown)


ChrisRackauckas:
> Moreover, I was wondering whether it might be reasonable to allow other initial vectors (for the autodifferentiation setup; currently it always creates a Vector{Float64}) and also custom chunk sizes, maybe as keyword arguments?

Sure.

>And regarding differentiation: I'm not up to date since I'm quite busy at the moment - but didn't you want to get rid of Calculus dependencies? Just noticed that the finite differentiation alternative still uses it here.

Yeah, using DiffEqDiffTools would be much better.

ChrisRackauckas:
LGTM but there were some odd test failures. I am not sure what to make of them.

devmotion:
On Linux `libnlopt0` cannot be installed somehow, but I have no clue why. The error on Win64 occurred before, e.g. on Win32 
https://ci.appveyor.com/project/ChrisRackauckas/diffeqparamestim-jl/build/1.0.186/job/097x7w9nbrca8vt6
and
https://ci.appveyor.com/project/ChrisRackauckas/diffeqparamestim-jl/build/1.0.168/job/actc6nqk678ewabi
It is always the same assertion in LineSearches that fails.

ChrisRackauckas:
Since Mac passed I'll merge, but it's really weird that it's so non-deterministic.

PR #61, opened 2018-01-11
update DiffEqParamEstim to latest

finmod:


ChrisRackauckas:
This would make a good PR to DiffEqBenchmarks.jl

https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl

We might as well add a parameter estimation section. 

>For the longer sample, all the algorithms fail. THIS IS A FAILURE OF THE NONLINEAR LEAST SQUARES ESTIMATOR L2LOSS.

What error estimator would you suggest then? You can add a custom loss function or regularization.

PR #71, opened 2018-03-29
Added first difference likelihood

Vaibhavdixit02:
This is an intial version and supports only Univariate distributions. I have kept it separate because I was getting quite poor accuracy in the test case so I wanted a review first before adding further things. @ChrisRackauckas

### DIFF
--- a/src/cost_functions.jl
+++ b/src/cost_functions.jl
@@ -70,36 +70,54 @@ function (f::L2Loss)(sol::AbstractMonteCarloSolution)
   mean(f.(sol.u))
 end
 
-struct LogLikeLoss{T,D} <: DECostFunction
+struct LogLikeLoss{T,D,F} <: DECostFunction
   t::T
   distributions::D
+  first_diff::F
 end
+LogLikeLoss(t,distributions) = LogLikeLoss(t,distributions,nothing)
 
 function (f::LogLikeLoss)(sol::DESolution)
-  distributions = f.distributions
-  fill_length = length(f.t)-length(sol)
-  for i in 1:fill_length
-    push!(sol.u,fill(Inf,size(sol[1])))
-  end
+  if f.first_diff == nothing
+    distributions = f.distributions
+    fill_length = length(f.t)-length(sol)
+    for i in 1:fill_length
+      push!(sol.u,fill(Inf,size(sol[1])))
+    end
 
-  ll = 0.0
+    ll = 0.0
 
-  if eltype(distributions) <: UnivariateDistribution
-    for j in 1:length(f.t), i in 1:length(sol[1][1])
+    if eltype(distributions) <: UnivariateDistribution
+      for j in 1:length(f.t), i in 1:length(sol[1][1])
       # i is the number of time points
       # j is the size of the system
       # corresponds to distributions[i,j]
-      ll -= logpdf(f.distributions[i,j],sol[i,j])
-    end
-  else # MultivariateDistribution
-    for j in 1:length(f.t), i in 1:length(sol[1][1])
+        ll -= logpdf(f.distributions[i,j],sol[i,j])
+      end
+    else # MultivariateDistribution
+      for j in 1:length(f.t), i in 1:length(sol[1][1])
       # i is the number of time points
       # j is the size of the system
       # corresponds to distributions[i,j]
-      ll -= logpdf(f.distributions[i],sol[i])
+        ll -= logpdf(f.distributions[i],sol[i])
+      end
     end
+    ll
+  else
+    distributions = f.distributions
+    diff_data = sol.u[2:end] - sol.u[1:end-1]
+    fill_length = length(f.t)-length(diff_data)
+    for i in 1:fill_length
+      push!(diff_data,fill(Inf,size(sol[1])))
+    end  
+    ll = 0.0
+    if eltype(distributions) <: UnivariateDistribution
+      for j in 1:length(f.t)-1, i in 1:length(diff_data[1])
+        ll -= logpdf(f.distributions[j,i],diff_data[j][i])
+      end
+    end
+    ll
   end
-  ll
 end
 
 function (f::LogLikeLoss)(sol::AbstractMonteCarloSolution)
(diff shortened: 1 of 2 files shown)


src/build_loss_objective.jl
```diff
@@ -8,7 +8,7 @@ end
 (f::DiffEqObjective)(x) = f.cost_function(x)
 (f::DiffEqObjective)(x,y) = f.cost_function2(x,y)
 
-function build_loss_objective(prob::DEProblem,alg,loss,regularization=nothing;mpg_autodiff = false,
+function build_loss_objective(prob::DEProblem,alg,loss,diff_likelihood=false,regularization=nothing;mpg_autodiff = false,
```
ChrisRackauckas: this should be part of the loss construction

src/cost_functions.jl
```diff
@@ -130,3 +130,19 @@ function (f::LogLikeLoss)(sol::AbstractMonteCarloSolution)
...
+  distributions = f.distributions
+  diff_data = sol.u[2:end] - sol.u[1:end-1]
+  fill_length = length(f.t)-length(diff_data)
+  for i in 1:fill_length
+    push!(diff_data,fill(Inf,size(sol[2]-sol[1])))
+  end  
+  ll = 0.0
+  if eltype(distributions) <: UnivariateDistribution
+    for j in 1:length(f.t)-1, i in 1:length(diff_data[1][1])
+      ll -= logpdf(f.distributions[j,i],diff_data[j][i])
```
ChrisRackauckas: this is row order

test/likelihood.jl
```diff
@@ -26,3 +26,27 @@ bound1 = Tuple{Float64, Float64}[(0.5, 5),(0.5, 5)]
...
+prob1 = ODEProblem(pf_func,u0,tspan,p)
+sol = solve(prob1,Tsit5())
+
+t = collect(linspace(0,10,200))
+function generate_data(sol,t)
+  randomized = VectorOfArray([(sol(t[i]) + .01randn(2)) for i in 1:length(t)])
+  data = convert(Array,randomized)
+end
+aggregate_data = convert(Array,VectorOfArray([generate_data(sol,t) for i in 1:100])) 
+distributions = [fit_mle(Normal,aggregate_data[i,j,:]-aggregate_data[i,j-1,:]) for j in 2:200, i in 1:2 ]
```
ChrisRackauckas: Make sure these distributions look correct.

src/build_loss_objective.jl
```diff
@@ -24,11 +24,17 @@ function build_loss_objective(prob::DEProblem,alg,loss,regularization=nothing;mp
     else
       sol = solve(tmp_prob,alg;kwargs...)
     end
+    
+    if diff_likelihood = true
+      loss_val = loss(sol,diff_likelihood=diff_likelihood)
+    else
+      loss_val = loss(sol)
+    end
```
ChrisRackauckas: This isn't needed if the boolean is in the likelihood

src/cost_functions.jl
```diff
@@ -72,34 +72,56 @@ end
...
-  end
+  if f.diff_distributions == nothing
+    distributions = f.data_distributions
+    fill_length = length(f.t)-length(sol)
+    for i in 1:fill_length
+      push!(sol.u,fill(Inf,size(sol[1])))
+    end
 
-  ll = 0.0
+    ll = 0.0
```
ChrisRackauckas: you want to initialize the `ll` to zero only one in order to sum them.

src/cost_functions.jl
```diff
@@ -72,34 +72,56 @@ end
...
+      for j in 1:length(f.t), i in 1:length(sol[1][1])
       # i is the number of time points
       # j is the size of the system
       # corresponds to distributions[i,j]
-      ll -= logpdf(f.distributions[i],sol[i])
+        ll -= logpdf(distributions[i],sol[i])
+      end
     end
+    ll
+  else
```
ChrisRackauckas: Since the `if` covers both, it won't put the two together and instead will only use one of the likelihoods.

src/cost_functions.jl
```diff
@@ -72,34 +72,56 @@ end
...
+    distributions = f.diff_distributions
+    diff_data = sol.u[2:end] - sol.u[1:end-1]
+    fill_length = length(f.t)-length(diff_data)
+    for i in 1:fill_length
+      push!(diff_data,fill(Inf,size(sol[1])))
+    end  
+    ll = 0.0
+    if eltype(distributions) <: UnivariateDistribution
```
ChrisRackauckas: this needs a multivariate distribution path

src/cost_functions.jl
```diff
@@ -72,34 +72,56 @@ end
...
+      for j in 1:length(f.t)-1, i in 1:length(diff_data[1])
+        ll -= logpdf(distributions[j,i],diff_data[j][i])
+      end
+    end
+    f1 = LogLikeLoss(f.t,f.data_distributions)
+    ll += f.weight*f1(sol)
+    ll
   end
-  ll
```
ChrisRackauckas: it should have a single return.

src/cost_functions.jl
Vaibhavdixit02: Yeah haven't covered that yet.

src/cost_functions.jl
```diff
@@ -72,34 +72,56 @@ end
...
+    for i in 1:fill_length
+      push!(diff_data,fill(Inf,size(sol[1])))
+    end  
+    ll = 0.0
+    if eltype(distributions) <: UnivariateDistribution
+      for j in 1:length(f.t)-1, i in 1:length(diff_data[1])
+        ll -= logpdf(distributions[j,i],diff_data[j][i])
+      end
+    end
+    f1 = LogLikeLoss(f.t,f.data_distributions)
```
Vaibhavdixit02: @ChrisRackauckas I think you missed these lines

src/cost_functions.jl
```diff
@@ -72,33 +72,56 @@ end
...
+    if eltype(distributions) <: UnivariateDistribution
+      for j in 1:length(f.t)-1, i in 1:length(diff_data[1])
+        ll -= logpdf(distributions[j,i],diff_data[j][i])
+      end
+    else
+      for j in 1:length(f.t)-1
+        ll -= logpdf(distributions[j],diff_data[j])
+      end
     end
+    ll += f.weight*f1(sol)
```
ChrisRackauckas: This can be deleted now?

PR #73, opened 2018-04-06
Added first difference likelihood for Monte Carlo method

Vaibhavdixit02:


src/cost_functions.jl
```diff
@@ -141,16 +141,34 @@ function (f::LogLikeLoss)(sol::AbstractMonteCarloSolution)
       # j is the size of the system
       # corresponds to distributions[i,j]
       vals = [s[i,j] for s in sol]
-      ll -= loglikelihood(f.distributions[i,j],vals)
+      ll -= loglikelihood(distributions[i,j],vals)
     end
   else
-    for j in 1:length(f.t)
+    for j in 1:length(f.t), i in 1:length(sol[1][1])
```
ChrisRackauckas: why this?

Vaibhavdixit02: Yeah, I believe it is a case of pasting in the wrong place.

PR #75, opened 2018-04-09
Added first difference to L2Loss and changes in Likelihood tests

Vaibhavdixit02:


src/cost_functions.jl
```diff
@@ -54,17 +58,33 @@ function (f::L2Loss)(sol::DESolution)
       for j in 1:length(sol[i])
         sumsq +=(data[j,i] - sol[j,i])^2
       end
+      if diff_data != nothing
+        for j in 2:length(sol[i])
+          sumsq +=(diff_data[j-1,i] - sol[j,i] + sol[j-1,i])^2
```
ChrisRackauckas: You don't actually need the `diff_data` for this one. Instead it's just `(data[j,i]-data[j,i-1] - sol[j,i] - sol[j,i-1])^2`. Notice the indexing by `i` too: the first column is through the solution at time point `i`.

test/likelihood.jl
```diff
@@ -10,100 +10,58 @@ tspan = (0.0,10.0)
...
+
 
+obj = build_loss_objective(prob1,Tsit5(),L2Loss(t,aggregate_data),
+                                     maxiters=10000,verbose=false)
+bound1 = Tuple{Float64, Float64}[(0.5, 5),(0.5, 5)]
+result = bboptimize(obj;SearchRange = bound1, MaxSteps = 11e3)
 @test result.archive_output.best_candidate ≈ [1.5,1.0] atol = 1e-1
+
+
+aggregate_data_diff = aggregate_data[:,1:end-1,:] - aggregate_data[:,2:end,:]
```
ChrisRackauckas: the user shouldn't need to generate this

ChrisRackauckas: Oh, this is the cause of that warning. `aggregate_data_diff = aggregate_data[:,1:end-1,:] - aggregate_data[:,2:end,:] ` that's a 3D matrix and in the cost function you're indexing it in 2D.

src/cost_functions.jl
Vaibhavdixit02: In that case, we always use the first difference too?

src/cost_functions.jl
```diff
@@ -54,17 +58,33 @@ function (f::L2Loss)(sol::DESolution)
       for j in 1:length(sol[i])
         sumsq +=(data[j,i] - sol[j,i])^2
       end
+      if diff_data != nothing
```
ChrisRackauckas: since we don't need this data, we can just do this off of whether a weight is given.

src/cost_functions.jl
ChrisRackauckas: We can do it based on whether there's a weight. But if the benchmarks show that it is good to use, it's probably always good to just do it then. The total cost should be minimal in comparison to solving the ODE itself, so anything that improves the estimate is good.

src/cost_functions.jl
Vaibhavdixit02: But it then requires user to always give weights incase they want to use the first difference too, which probably will not be the case mostly?

Vaibhavdixit02: I mean between giving weights and not giving weights, in my opinion not having weights would be the more common one?

src/cost_functions.jl
Vaibhavdixit02: Doing it always would probably be better

src/cost_functions.jl
ChrisRackauckas: Yeah, I'm not sure. Maybe just set the weight to 1 by default and just always do the calculation unless it's zero.

src/cost_functions.jl
```diff
@@ -54,17 +56,29 @@ function (f::L2Loss)(sol::DESolution)
       for j in 1:length(sol[i])
         sumsq +=(data[j,i] - sol[j,i])^2
       end
+      if diff_weight != nothing
+        for j in 2:length(sol[i])
+          sumsq += diff_weight[j-1,i]*((data[j,i] - data[j-1,i] - sol[j,i] + sol[j-1,i])^2)
```
ChrisRackauckas: This is differencing across state, not time.

ChrisRackauckas:
The issue the tests are having is that the aggregate data for maximum likelihood is different from the data the the L2Loss uses. The aggregate data is of `N` trajectories and uses the distribution over that data. The L2Loss wants a single trajectory to match. This test should instead use the `tests_on_odes` part, adding a new file in there that uses the `data` but with different weights on the two parts.

PR #76, opened 2018-04-13
multiple_shooting_objective implemented

Vaibhavdixit02:


src/multiple_shooting_objective.jl
```diff
@@ -0,0 +1,56 @@
...
+    sol = []
+    loss_val = 0
+    for i in 1:2:N
+        tmp_prob = remake(prob;u0=p[i:i+1],p=[N+1:N+2])
+        if typeof(loss) <: Union{CostVData,L2Loss,LogLikeLoss}
+          # println("huihui1")
+          push!(sol,solve(tmp_prob,alg;saveat=time_dur,save_everystep=false,dense=false,kwargs...))
+          # println("huihui2")
+          if (i+1)*time_len < length(loss.t)
+            time_dur = loss.t[i*time_len:(i+1)*time_len]
```
ChrisRackauckas: this should be before you use it in saveat?

src/multiple_shooting_objective.jl
```diff
@@ -0,0 +1,56 @@
...
+          else
+            time_dur = loss.t[i*time_len:Int(length(loss.t))]
+          end
+        else
+          push!(sol,solve(tmp_prob,alg;kwargs...))
+        end
+
+        if regularization == nothing
+            loss_val += loss(sol[i])
```
ChrisRackauckas: oh, somehow the loss function has to know the time range.

src/multiple_shooting_objective.jl
Vaibhavdixit02: Have initialized before and then updating it after using it in `saveat`

src/multiple_shooting_objective.jl
```diff
@@ -1,20 +1,30 @@
 export multiple_shooting_objective
+
+function generate_loss_func(loss,t,i)
+  new_loss = nothing
+  if typeof(loss)::L2Loss
+    new_loss = L2Loss(t,loss.data[:,i:length(t),:])
```
ChrisRackauckas: This data shouldn't be 3d?

ChrisRackauckas:
Tests failing due to indexing error.

src/multiple_shooting_objective.jl
```diff
@@ -71,4 +76,23 @@ function multiple_shooting_objective(prob::DEProblem,alg,loss,regularization=not
...
+end
+
+function merge_solutions(prob::DEProblem,alg,t,final_params)
+  N = length(final_params)-length(prob.p)
+  time_len = Int(floor(length(t)/N))
+  time_dur = t[1:time_len]
+  sol = []
+  for i in 1:length(prob.u0):N
+      tmp_prob = remake(prob;u0=final_params[i:i+length(prob.u0)-1],p=final_params[N+1:N+length(prob.p)],tspan=(time_dur[1],time_dur[end]))
+      push!(sol,solve(tmp_prob,alg;saveat=time_dur,save_everystep=false,dense=false))
```
ChrisRackauckas: instead of the merged solution, I would want this `sol` array out.

src/multiple_shooting_objective.jl
```diff
@@ -0,0 +1,102 @@
...
+        j = j+1
+    end
+    u = [i for j in 1:length(sol) for i in sol[j].u]
+    t = [i for j in 1:length(sol) for i in sol[j].t]
+    sol_loss = Merged_Solution(u,t,sol)
+    time_dur = loss.t[1:time_len]
+    for i in 2:length(sol)
+      loss_val += discontinuity_weight*sum(sol[i][1] - sol[i-1][end])^2
+    end
+    sol = sol = build_solution(prob,alg,sol_loss.t,sol_loss.u)
```
ChrisRackauckas: two `=`

ChrisRackauckas: can you move this up to line 52?

src/multiple_shooting_objective.jl
```diff
@@ -28,16 +28,17 @@ function multiple_shooting_objective(prob::DEProblem,alg,loss,regularization=not
   cost_function = function (p)
     N = length(p)-length(prob.p)
     time_len = Int(floor((length(loss.t)*length(prob.u0))/N))
-    time_dur = loss.t[1:time_len]
+    time_dur = loss.t[1:time_len+1]
     sol = []
     loss_val = 0
     j = 1
+    data_my = solve(prob,alg,abstol=1e-12,reltol=1e-12;saveat=loss.t,save_everystep=false,dense=false,kwargs...)
```
ChrisRackauckas: `abstol=1e-12,reltol=1e-12` just have that passed in via the kwargs.

PR #77, opened 2018-04-14
Remove warnings arising in L2Loss

Vaibhavdixit02:


src/cost_functions.jl
```diff
@@ -84,9 +84,7 @@ function (f::L2Loss)(sol::DESolution)
   end
   sumsq
 end
-L2Loss(t,data;data_weight=nothing,differ_weight=nothing) = L2Loss(t,data,data_weight,differ_weight)
-L2Loss(t,data,differ_weight=1;data_weight=nothing) = L2Loss(t,data,differ_weight,data_weight)
-# L2Loss(t,data,differ_data,differ_weight;weight=nothing) = L2Loss(t,data,differ_data,weight,differ_weight)
+L2Loss(t,data;data_weight=1,differ_weight=nothing) = L2Loss(t,data,data_weight,differ_weight)
```
ChrisRackauckas: I would default to a float here since it gives better type stability.

ChrisRackauckas:
Test failure

src/cost_functions.jl
```diff
@@ -84,9 +84,8 @@ function (f::L2Loss)(sol::DESolution)
   end
   sumsq
 end
-L2Loss(t,data;data_weight=nothing,differ_weight=nothing) = L2Loss(t,data,data_weight,differ_weight)
-L2Loss(t,data,differ_weight=1;data_weight=nothing) = L2Loss(t,data,differ_weight,data_weight)
-# L2Loss(t,data,differ_data,differ_weight;weight=nothing) = L2Loss(t,data,differ_data,weight,differ_weight)
+L2Loss(t,data) = L2Loss(t,data,nothing,nothing)
```
ChrisRackauckas: that's not necessary.

PR #81, opened 2018-05-10
Extend l2loss_test.jl to cover more cases

Vaibhavdixit02:


ChrisRackauckas:
Why would different data matter? Our methods shouldn't modify the data.

Vaibhavdixit02:
Well the test seems to be failing because of some bounds error so I was just checking if there is some difference in data.

PR #84, opened 2018-05-14
Added priors as argument to cost function and objective function

Vaibhavdixit02:


src/cost_functions.jl
```diff
@@ -1,15 +1,30 @@
...
 Regularization(λ) = Regularization{typeof(λ),typeof(L2Penalty())}(λ,L2Penalty())
 
+struct Prior
+  prior
+end 
 function (f::Regularization)(p)
   f.λ*value(f.penalty, p)
 end
 
+function (f::Prior)(p)
```
ChrisRackauckas: This would be easier as just a function call

src/build_loss_objective.jl
```diff
@@ -25,10 +25,16 @@ function build_loss_objective(prob::DEProblem,alg,loss,regularization=nothing;mp
       sol = solve(tmp_prob,alg;kwargs...)
     end
 
-    if regularization == nothing
+    if regularization == nothing && prior == nothing
       loss_val = loss(sol)
-    else
+    elseif prior == nothing
       loss_val = loss(sol) + regularization(p)
```
ChrisRackauckas: I think just adding to a loss val would really clean up this code

Vaibhavdixit02: Didn't quite understand what you mean here?

src/build_loss_objective.jl
```diff
@@ -25,11 +25,15 @@ function build_loss_objective(prob::DEProblem,alg,loss,regularization=nothing;mp
...
-      loss_val = loss(sol)
-    else
-      loss_val = loss(sol) + regularization(p)
+    loss_val = loss(sol)
+    
+    if prior != nothing
+      loss_val += prior_loss(prior,p)
+    end
+    if regularization != nothing
+      loss_val = loss(sol) +
```
ChrisRackauckas: mistake? Should be `+=` then the regularization?

src/multiple_shooting_objective.jl
```diff
@@ -53,7 +53,11 @@ function multiple_shooting_objective(prob::DEProblem,alg,loss,regularization=not
     push!(t,sol[end].t[end])
     sol_loss = Merged_Solution(u,t,sol)
     sol_new = build_solution(prob,alg,loss.t,sol_loss.u)
-    loss_val += loss(sol_new)
+    if prior == nothing
+      loss_val += loss(sol_new)
+    else
+      loss_val += loss(sol_new) + prior_loss(prior,p)
+    end
```
ChrisRackauckas: let's add regularization while we're at it.

src/cost_functions.jl
```diff
@@ -10,6 +10,18 @@ function (f::Regularization)(p)
...
 end
 
+function prior_loss(prior,p)
+  ll = 0.0
+  if eltype(prior) <: UnivariateDistribution
+    for i in 1:length(prior)
+      ll -= logpdf(prior[i],p[i])
+    end
+  else
+    ll -= logpdf(prior[1],p)
```
ChrisRackauckas: if it's just a multivariate distribution, does it need to be indexed?

Vaibhavdixit02: Taking array as input to maintain uniformity?

PR #85, opened 2018-07-17
Upgrade to 0.7

Vaibhavdixit02:


Vaibhavdixit02:
@ChrisRackauckas https://github.com/JuliaDiffEq/DiffEqParamEstim.jl/blob/master/src/cost_functions.jl#L3 This errors that `DECostFunction` is not declared, is there some syntax change in this case in 0.7?

ChrisRackauckas:
>This errors that DECostFunction is not declared, is there some syntax change in this case in 0.7 that femtocleaner can not catch?

Non-user facing parts of DiffEqBase.jl are no longer exported, so that needs to be DiffEqBase.DECostFunction.

Vaibhavdixit02:
@ChrisRackauckas this stack trace looks similar to DiffEqSensitivity's, it is pointing at `problem_new_parameters`, I think it can be merged now?

ChrisRackauckas:
`DiffEqBase.problem_new_parameters`?

Vaibhavdixit02:
That isn't the problem, https://travis-ci.org/JuliaDiffEq/DiffEqParamEstim.jl/jobs/405096216#L2606 some change in `ODEProblem`, kwargs `callback` and `mass_matrix` seem to be the problem.

ChrisRackauckas:
Oh yes. Anyways, `problem_new_parameters(prob,p)` should instead be rewritten to just be `remake(prob;p=p)` now that we have more generic problem creation functions.

ChrisRackauckas:
straight remake won't work there. You want the function `(prob,p) -> remake(prob;p=p)`

Vaibhavdixit02:
I decided to use the most straightforward way to deal with the issue that was causing the error. Now we are held up Optim.jl it seems from the error. I checked their repo and they seem to be in the process of updating so will keep following them and resume the work this once they release the updated version.

Vaibhavdixit02:
Oh and I somehow messed up and pushed some commits to the master branch, you might want to do something about that, I couldn't. But we can also let it be as those commits will be merged with this PR anyways. Your call @ChrisRackauckas.

ChrisRackauckas:
Don't comment out tests. Mark them `@test_broken` so they don't get lost.

I don't think we should workaround the vector thing. And actually, I think the "simple" integration with LossFunctions.jl should just go away: it's slower and we have a more general interface anyways. I'll take a stab at this soon enough, but if you could revert any test commenting that would be helpful.

test/tests_on_odes/blackboxoptim_test.jl
```diff
@@ -1,20 +1,20 @@
...
+# bound1 = Tuple{Float64, Float64}[(1, 2)]
+# result = bboptimize(cost_function;SearchRange = bound1, MaxSteps = 11e3)
+# @test_broken result.archive_output.best_candidate[1] ≈ 1.5 atol=3e-1
 
-cost_function = build_loss_objective(prob2,Tsit5(),L2Loss(t,data),
-                                     maxiters=10000,verbose=false)
-bound2 = Tuple{Float64, Float64}[(1, 2),(2, 4)]
-result = bboptimize(cost_function;SearchRange = bound2, MaxSteps = 11e3)
-@test_broken result.archive_output.best_candidate ≈ [1.5;3.0] atol=3e-1
+# cost_function = build_loss_objective(prob2,Tsit5(),L2Loss(t,data),
```
ChrisRackauckas: `@test_broken` the blocks.

Vaibhavdixit02: I didn't want to import BlackBoxOptim at all. It's failing pre-compilation so seemed like a good idea to look over these completely as it stops the other errors down the line that can be fixed right now.

ChrisRackauckas: Yes, comment out the import and the `@test_broken` the usage areas.

PR #92, opened 2018-08-15
 v0.7 upgrades

Vaibhavdixit02:
So the tests pass for me locally now. With some points because of some (point 1) of which the tests will not pass on Travis:

1. BlackBoxOptim.jl needs a new release tag and I have opened an issue for the same here (it has been upgraded to 0.7 but not a new release so I used their master branch in running the tests locally).

2. https://github.com/Vaibhavdixit02/DiffEqParamEstim.jl/blob/master/test/tests_on_odes/nlopt_test.jl#L41 this test behaves weirdly, putting it in `@test_broken` gives error that it is a passing test but it fails on `@test` with the previous tolerance of `1e-3`, I suspect it has something to do with the scoping rules and carrying over of old values in the loop (maybe?). 

3. Two stage method needs a lot of upgrades and there is already an issue for that.

4. `┌ Warning: Deprecated syntax `implicit assignment to global variable `obj``.
│ Use `global obj` instead.` I remember discussing this deprecation warning with you in slack (couldn't find the message), this is supposed to be ignored in 0.7, right?

### DIFF
--- a/test/tests_on_odes/two_stage_method_test.jl
+++ b/test/tests_on_odes/two_stage_method_test.jl
@@ -1,27 +1,27 @@
 println("Use Two Stage Method to fit the parameter")
 
 @test_broken begin
-  cost_function = two_stage_method(prob1,t,data)
-  result = Optim.optimize(cost_function, 0.0, 10.0)
-  @test result.minimizer[1] ≈ 1.5 atol=3e-1
-  cost_function = two_stage_method(prob2,t,data)
-  result = Optim.optimize(cost_function, [1.0,2.5], Optim.BFGS())
-  @test result.minimizer ≈ [1.5;3.0] atol=3e-1
-  cost_function = two_stage_method(prob3,t,data)
-  result = Optim.optimize(cost_function, [1.3,0.8,2.8,1.2], Optim.BFGS())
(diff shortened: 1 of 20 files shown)


Vaibhavdixit02:
@ChrisRackauckas can you take a look at the Travis error here?

ChrisRackauckas:
Looks like this conflicts with master?

test/tests_on_odes/two_stage_method_test.jl
```diff
@@ -1,29 +1,27 @@
...
-      obj = two_stage_method(prob3,t,data; mpg_autodiff = autodiff)
-      opt = Opt(:LD_LBFGS, 4)
-      min_objective!(opt, obj.cost_function2)
-      (minf,minx,ret) = NLopt.optimize(opt, [1.3,0.8,2.8,1.2])
-      @test minx ≈ [1.5;1.0;3.0;1.0] atol=5e-1
-    end
-end
\ No newline at end of file
+
+@test_broken begin
```
ChrisRackauckas: why is this broken? Optim should be working?

ChrisRackauckas:
Is this no longer WIP then?

Vaibhavdixit02:
Yeah

PR #98, opened 2018-12-28
Use Adjoint for SA gradients

Vaibhavdixit02:


src/build_loss_objective.jl
```diff
@@ -49,9 +65,18 @@ function build_loss_objective(prob::DiffEqBase.DEProblem,alg,loss,regularization
   if mpg_autodiff
     gcfg = ForwardDiff.GradientConfig(cost_function, autodiff_prototype, autodiff_chunk)
     g! = (x, out) -> ForwardDiff.gradient!(out, cost_function, x, gcfg)
+  elseif lsa_gradient 
+    function g!(x,out)
+      sol_,sens = diffeq_sen_full(prob.f,prob.u0,prob.tspan,x,loss.t;alg=alg)
+      l2lossgradient!(out,sol_,loss.data,sens,length(prob.p))
```
ChrisRackauckas: this is only for L2loss. Throw an error if it's not used?

src/build_loss_objective.jl
```diff
@@ -49,9 +65,18 @@ function build_loss_objective(prob::DiffEqBase.DEProblem,alg,loss,regularization
...
+    end
+  elseif adjsa_gradient
+    g! = (x,out) -> diffeq_sen_l2!(out,prob.f,prob.u0,prob.tspan,x,loss.t,loss.data,alg)
   else
     g! = (x, out) -> Calculus.finite_difference!(cost_function,x,out,:central)
   end
+
+
   cost_function2 = function (p,grad)
```
ChrisRackauckas: you need to change this for the adjoint to reduce a solve.

src/build_loss_objective.jl
```diff
@@ -5,6 +5,22 @@ struct DiffEqObjective{F,F2} <: Function
...
 end
 
+function diffeq_sen_full(f, u0, tspan, p, t; alg=Tsit5(), kwargs...)
+  prob = ODELocalSensitivityProblem(f,u0,tspan,p)
+  sol = solve(prob,alg;kwargs...)(t)
+  nvar = length(u0)
+  sol[1:nvar,:], [sol[i*nvar+1:i*nvar+nvar,:] for i in 1:length(p)]
+end
+
+function diffeq_sen_l2!(res, df, u0, tspan, p, t, data, alg=Tsit5();kwargs...)
```
ChrisRackauckas: no default algorithm needed here.

src/build_loss_objective.jl
Vaibhavdixit02: Can you explain what you mean?

Vaibhavdixit02: We could finish this up once you can respond here?

src/build_loss_objective.jl
```diff
@@ -5,6 +5,22 @@ struct DiffEqObjective{F,F2} <: Function
   cost_function2::F2
 end
 
+function diffeq_sen_full(f, u0, tspan, p, t; alg=Tsit5(), kwargs...)
```
ChrisRackauckas: this isn't used.

src/build_loss_objective.jl
ChrisRackauckas: https://github.com/JuliaOpt/NLopt.jl#tutorial notice that the gradient and the value computation are mixed together here. In the adjoint calculation, the forward solution is required, so that can be used instead of re-solving.

src/cost_functions.jl
```diff
@@ -190,3 +190,14 @@ function (f::LogLikeLoss)(sol::DiffEqBase.AbstractMonteCarloSolution)
 
   ll
 end
+
+function l2lossgradient!(grad,sol,data,sensitivities,num_p)
```
ChrisRackauckas: this assumes the ODE is a vector ODE?

src/build_loss_objective.jl
Vaibhavdixit02: https://github.com/JuliaDiffEq/DiffEqParamEstim.jl/pull/98/files/51946011756285f36f80c172712a5a0ae6c86460#diff-5d1e39f5dbff55f43e1133fa9262904eR71?

src/build_loss_objective.jl
Vaibhavdixit02: That looks NLopt specific though? https://github.com/JuliaDiffEq/DiffEqParamEstim.jl/pull/98/files/51946011756285f36f80c172712a5a0ae6c86460#diff-2fed90081d9e12363b34887426589f36R43 optimisers require a separate function call for the the loss and for gradient evaluation generally?

ChrisRackauckas: Look at Optim's `fg!`

src/build_loss_objective.jl
```diff
@@ -49,9 +65,22 @@ function build_loss_objective(prob::DiffEqBase.DEProblem,alg,loss,regularization
   if mpg_autodiff
     gcfg = ForwardDiff.GradientConfig(cost_function, autodiff_prototype, autodiff_chunk)
     g! = (x, out) -> ForwardDiff.gradient!(out, cost_function, x, gcfg)
+  elseif lsa_gradient
+    if typeof(loss) <: L2Loss
+      function g!(x,out)
+        sol_,sens = diffeq_sen_full(prob.f,prob.u0,prob.tspan,x,loss.t;alg=alg)
```
ChrisRackauckas: why is this using the forward sensitivity analysis?

src/build_loss_objective.jl
ChrisRackauckas: it shouldn't be used?

src/build_loss_objective.jl
Vaibhavdixit02: We had a conversation the other day when we decided that we will support this as well as the adjoint sensitivity analysis method just in case user in interested. I don't see why not leave it there (it does work and is tested) incase someone would like to use it for some possible use case?

ChrisRackauckas: ahh yes, lsa instead of adjlsa. Change the name to flsa for forward.

src/build_loss_objective.jl
ChrisRackauckas: will be addressed in a different PR.

ChrisRackauckas:
Still some comments to address.

src/build_loss_objective.jl
ChrisRackauckas: no default algorithm needed here?

Vaibhavdixit02:
LGTM?

src/cost_functions.jl
ChrisRackauckas: Open an issue for this.

Issue #74, opened 2018-04-08
Spline or Hermite derivative term for the L2 loss


ChrisRackauckas:
We can add a term where a cubic spline is used to approximate the derivative which is then checked against `f(sol,p,t)`. The minima of this would now only be approximate, but maybe this could help enlarge the true basin.

PR #105, opened 2019-02-07
 Added spline derivative term for the L2 loss (#74)

yakovlev31:


src/cost_functions.jl
```diff
@@ -77,20 +80,41 @@ function (f::L2Loss)(sol::DiffEqBase.DESolution)
...
+  if colloc_grad != nothing
+    dudt = zeros(size(colloc_grad))
+    tmp = zeros(size(colloc_grad)[1])
+    for i = 1:size(colloc_grad)[2]
+      sol.prob.f.f(tmp, sol.u[i], sol.prob.p, sol.t)
+      for j = 1:length(tmp)
+        dudt[j,i] = tmp[j]
+      end
+    end
+    sumsq += sum(sum((dudt - colloc_grad).^2, dims=2), dims=1)[1]
```
ChrisRackauckas: `sum(abs2,x - y for (x,y) in zip(dudt,colloc_grad))` should work without the allocations.

src/cost_functions.jl
```diff
@@ -77,20 +80,41 @@ function (f::L2Loss)(sol::DiffEqBase.DESolution)
       end
     end
   end
+  if colloc_grad != nothing
+    dudt = zeros(size(colloc_grad))
+    tmp = zeros(size(colloc_grad)[1])
+    for i = 1:size(colloc_grad)[2]
+      sol.prob.f.f(tmp, sol.u[i], sol.prob.p, sol.t)
```
ChrisRackauckas: `sol.prob.f.f(@view(dudt[:,i]), sol.u[i], sol.prob.p, sol.t)`

ChrisRackauckas: then `tmp` isn't needed.

src/cost_functions.jl
```diff
@@ -77,20 +80,41 @@ function (f::L2Loss)(sol::DiffEqBase.DESolution)
       end
     end
   end
+  if colloc_grad != nothing
+    dudt = zeros(size(colloc_grad))
```
ChrisRackauckas: This can be saved in the loss function rather than re-created each time.

yakovlev31:
Thanks! Fixed.

ChrisRackauckas:
The method looks good, but this needs to be added to the tests.

test/tests_on_odes/l2_colloc_grad_test.jl
```diff
@@ -0,0 +1,20 @@
...
+result = bboptimize(cost_function;SearchRange = bound1, MaxSteps = 11e3)
+@test result.archive_output.best_candidate[1] ≈ 1.5 atol=3e-1
+
+cost_function = build_loss_objective(prob2,Tsit5(),L2Loss(t,data,colloc_grad=colloc_grad(t, data)),
+                                     maxiters=10000,verbose=false)
+bound2 = Tuple{Float64, Float64}[(1, 2),(1, 4)]
+result = bboptimize(cost_function;SearchRange = bound2, MaxSteps = 11e3)
+@test result.archive_output.best_candidate ≈ [1.5;3.0] atol=3e-1
+
+cost_function = build_loss_objective(prob3,Tsit5(),L2Loss(t,data,colloc_grad=colloc_grad(t, data)),
```
ChrisRackauckas: I'd like to see one where the weight on the shooting part is set really small so that it's a unit test on the collocation part.

yakovlev31: The idea was that the collocation and weight parts don't intersect and if they work correctly separately they will work correctly together. Or do I miss something here?

ChrisRackauckas: I don't think you've tested the separately?

yakovlev31:
Weird...Works well on my laptop.

PR #138, opened 2020-09-07
reduce allocations by two_stage_method

baggepinnen:
This reduces the amount of allocations by about a factor 5 and runs at least some 30% faster


ChrisRackauckas:
great thanks!

PR #142, opened 2020-10-12
Enhancing multiple_shooting_objective

wg030:
I am currently testing the DiffEqParamEstim package and I figured that the `multiple_shooting_objective` contained some small misstakes. For example the `verbose_opt` did not work because the `verbose_steps` was missing.
Moreover the formula for discontinuity part of the loss seemed to be slightly incorrect.
I also tried to improve how the function deals with the `loss` so that the user can now provide his own function without creating it as `Union{L2Loss,LogLikeLoss}`.


ChrisRackauckas:
This looks spot on. Is there a good way to test it better than we did before to catch the issues on the discontinuities?

wg030:
I don't really know a good way for testing the discontinuities specifically. When I tested the code I also used the Lorenz Attractor.
Btw thanks for merging!

PR #143, opened 2020-10-13
Fix check of return code

devmotion:
Seems like the check of the return codes was incorrect.


devmotion:
BTW no that I have another look at the checks of the ensemble solutions, it seems they are also incorrect. Shouldn't they check if the return code of any of the solutions is neither `:Success` nor `:Terminated` instead of if there is a solution whose return code is not `:Success` and one whose return code is not `:Terminated` (but not necessarily the same solutions)?

Vaibhavdixit02:
@devmotion yes

PR #144, opened 2020-10-31
Create index.md

mkg33:
Add the documentation directory (content copied verbatim from DiffEqDocs.jl/parameter_esimation.md).


ChrisRackauckas:
Good start. I think we will want to split this up into a few different pages and such, but let's start with this.

mkg33:
Yeah, that's what I thought but I didn't know how big the respective sections would eventually get. It'll be easier to split the file logically once we've updated it.

PR #155, opened 2021-01-01
Rename prior arg to priors

Vaibhavdixit02:


ChrisRackauckas:
? is that how it was documented or something?

Vaibhavdixit02:
@ChrisRackauckas yeah that's how it was documented and used in the tests but the function signature had `prior` 🤦

PR #179, opened 2022-10-15
remove sqrt in losses

ArnoStrouwen:


Vaibhavdixit02 (commented):
Needs the test values updated

PR #180, opened 2022-10-15
remove duplicate code for collocation

ArnoStrouwen:
DiffEqFlux has copy pastes of some of the functions here.
They should be removed in one of the two.
https://github.com/SciML/DiffEqFlux.jl/issues/773

I would also suggest deprecating passing the kernel as a symbol, to make it more uniform with the rest of the SciML interfaces, e.g. `Tsit5()` is used, not `:Tsit5`.


Project.toml
```diff
@@ -7,6 +7,7 @@ version = "1.26.0"
 Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 Dierckx = "39dd38d3-220a-591b-8e3c-4c3a8c710a94"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
+DiffEqFlux = "aae7a2af-3d4f-5e19-a356-7da93b79d9d0"
```
Vaibhavdixit02, line 10: I don't think we should take this on as a dependency...

src/two_stage_method.jl
```diff
@@ -32,51 +34,19 @@ function decide_kernel(kernel::Symbol)
...
         return SigmoidKernel()
-    else
+    elseif kernel == :Silverman
         return SilvermanKernel()
+    else
+        return error("Kernel name not recognized")
     end
 end
 
-function construct_t1(t, tpoints)
```
Vaibhavdixit02, line 40: Why delete these lines? 🤔

src/two_stage_method.jl
```diff
@@ -102,23 +72,27 @@ get_chunksize(cs) = cs
...
                           loss_func = L2Loss, mpg_autodiff = false,
                           verbose = false, verbose_steps = 100,
                           autodiff_chunk = length(prob.p))
-    f = prob.f
-    kernel_function = decide_kernel(kernel)
-    estimated_derivative, estimated_solution = construct_estimated_solution_and_derivative!(data,
-                                                                                            kernel_function,
-                                                                                            tpoints)
+    if kernel isa Symbol
+        @warn "Passing kernels as Symbols will be deprecated"
```
Vaibhavdixit02, line 80: You can use @deprecate here

ArnoStrouwen: I wanted to do it like that at first, but I ran into issues because Julia cannot dispatch on a keyword argument, i.e. it is impossible to have two methods, one with kernel restricted to `Symbol` types and one with kernel restricted to `CollocationKernel`, and then `@deprecate` the `Symbol` method.

Project.toml
ChrisRackauckas: Definitely not taking that dependency here.

src/two_stage_method.jl
ChrisRackauckas: I don't think you can `@depreacate` here, it needs this special handling. But I also don't think we should be merging deprecations to this repo: it's somewhat in freeze mode because its interface does not extend well to more cases. In that case, major dependency changes and deprecations for an old repo not getting any major benefits is not a great feel. I think this PR should just hold until there's a good reason found for it.

Vaibhavdixit02: Yeah without kwarg handling it's not obvious, I think the way to do it with `@deprecate` would be 
(though it's not relevant anymore since we won't go ahead with the PR, but it might be useful later)

```
my_func(args...; kwarg_to_deprecate::Union{T1,T2}, kwargs...) = my_func(args..., kwarg_to_deprecate; kwargs...)

my_func(args..., kwarg_to_deprecate::T1; kwargs....) = do something...

@deprecate my_func(args..., kwarg_to_deprecate::T2; kwargs...) = do something....
```

PR #196, opened 2022-12-12
Refactor code to make it Optimization.jl compatible

Vaibhavdixit02:


src/build_loss_objective.jl
```diff
@@ -68,10 +64,7 @@ function build_loss_objective(prob::SciMLBase.AbstractSciMLProblem, alg, loss,
         loss_val
     end
 
-    if mpg_autodiff
-        gcfg = ForwardDiff.GradientConfig(cost_function, autodiff_prototype, autodiff_chunk)
-        g! = (x, out) -> ForwardDiff.gradient!(out, cost_function, x, gcfg)
-    elseif flsa_gradient
+    if flsa_gradient
```
ChrisRackauckas, line 67: this is also an autodiff option

src/build_loss_objective.jl
```diff
@@ -83,15 +76,17 @@ function build_loss_objective(prob::SciMLBase.AbstractSciMLProblem, alg, loss,
...
+    if flsa_gradient || adjsa_gradient
+        cost_function2 = function (p, grad)
+            if length(grad) > 0
+                g!(p, grad)
+            end
+            cost_function(p)
         end
-        cost_function(p)
+    else
+        cost_function2 = nothing
```
ChrisRackauckas, line 89: the cost_function2 can just in general be removed.

Vaibhavdixit02: Yeah even I thought that but wasn't sure if it would be too breaking

ChrisRackauckas: Go for broke: This should get a major release with this change anyways. Just make these functions that return objective functions into functions that return an OptimizationProblem or something of the sort.

Vaibhavdixit02: Okay cool

src/two_stage_method.jl
```diff
@@ -69,62 +69,31 @@ function construct_estimated_solution_and_derivative!(data, kernel, tpoints)
...
-    else
-        cost_function = construct_oop_cost_function(f, du, preview_est_sol,
-                                                    preview_est_deriv, tpoints)
+
+    function cost_function(p, nothing)
+        cost = zero(first(p))
+        for i in 1:length(preview_est_sol)
+            est_sol = preview_est_sol[i]
+            _du = f(est_sol, p, tpoints[i])
+            cost += sum(abs2, vec(preview_est_deriv[i]) .- vec(_du))
```
ChrisRackauckas, line 92: only oop?

Project.toml
```diff
@@ -8,14 +8,12 @@ Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
...
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
-ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 LsqFit = "2fda8390-95c7-5789-9bda-21331edee243"
 PenaltyFunctions = "06bb1623-fdd5-5ca2-a01c-88eae3ea319e"
 PreallocationTools = "d236fae5-4411-538c-8e31-a6e3d9e00b46"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
-SciMLSensitivity = "1ed8b502-d754-442c-8d5d-10ac956f44a1"
```
ChrisRackauckas, line 18: it might be good to keep a SciMLSensitivity because users just get a weird error telling them to add it if they use AutoZygote, and I think it won't be too uncommon to do that. 

That said, a better solution may be to make SciMLSensitivity be a weak dependency of DiffEqBase that is added and used whenever Zygote, ReverseDiff, or Tracker are used. @oscardssmith is that possible?

src/two_stage_method.jl
```diff
@@ -71,8 +69,9 @@ function construct_estimated_solution_and_derivative!(data, kernel, tpoints)
     estimated_solution = reduce(hcat, transpose.(last.(x)))
     estimated_derivative, estimated_solution
 end
+
 function construct_iip_cost_function(f, du, preview_est_sol, preview_est_deriv, tpoints)
-    function (p)
+    function (p, nothing)
```
ChrisRackauckas, line 74:
```suggestion
    function (p, _)
```

don't shadow

src/two_stage_method.jl
```diff
@@ -87,7 +86,7 @@ function construct_iip_cost_function(f, du, preview_est_sol, preview_est_deriv,
 end
 
 function construct_oop_cost_function(f, du, preview_est_sol, preview_est_deriv, tpoints)
-    function (p)
+    function (p, nothing)
```
ChrisRackauckas, line 89:
```suggestion
    function (p, _)
```

src/multiple_shooting_objective.jl
```diff
@@ -22,21 +22,12 @@ struct Merged_Solution{T1, T2, T3}
...
-                                     autodiff_chunk = mpg_autodiff ?
-                                                      ForwardDiff.Chunk(autodiff_prototype) :
-                                                      nothing,
                                      kwargs...)
-    if verbose_opt
-        count = 0 # keep track of # function evaluations
-    end
-
-    cost_function = function (p)
+    cost_function = function (p, _p = nothing)
```
ChrisRackauckas, line 30:
```suggestion
    cost_function = function (p, _)
```

src/build_loss_objective.jl
```diff
@@ -1,45 +1,12 @@
...
-                              autodiff_chunk = mpg_autodiff ?
-                                               ForwardDiff.Chunk(autodiff_prototype) :
-                                               nothing, flsa_gradient = false,
-                              adjsa_gradient = false,
                               kwargs...)
-    if verbose_opt
-        count = 0 # keep track of # function evaluations
-    end
-    cost_function = function (p)
+    cost_function = function (p, _p = nothing)
```
ChrisRackauckas, line 9:
```suggestion
    cost_function = function (p, _)
```

test/lorenz_test.jl
```diff
@@ -34,184 +34,91 @@ data = convert(Array, solve(prob, Euler(), tstops = t))
...
+# opt = Opt(:GN_ISRES, 3)
+# res = solve(optprob, opt)
 
 opt = Opt(:LN_BOBYQA, 3)
-lower_bounds!(opt, [9.0, 20.0, 2.0])
-upper_bounds!(opt, [11.0, 30.0, 3.0])
-min_objective!(opt, obj_short.cost_function2)
-xtol_rel!(opt, 1e-12)
-maxeval!(opt, 10000)
-@time (minf, minx, ret) = NLopt.optimize(opt, [9.0, 20.0, 2.0])
```
ChrisRackauckas, line 201: beautiful

ChrisRackauckas:
Let's merge but not tag until docs

Project.toml
oscardssmith: yeah that would work

PR #199, opened 2022-12-21
Update docs

Vaibhavdixit02:


ChrisRackauckas:
I rebased BTW.

Vaibhavdixit02:
Shit, I missed that

ChrisRackauckas:
I'll force over yours, and then let you finish it.

Vaibhavdixit02:
Thanks! Sorry about that

ChrisRackauckas:
You take it from here. I just jolted it into the standard form for all of our new docs. I am not sure if it all runs but it should be close.

Vaibhavdixit02:
Great 👍

ChrisRackauckas:
BTW, we should show AutoForwardDiff here since Zygote does not make sense on a small model like this. To showcase the Zygote choice, there should be another tutorial that's focused on a larger model.

ChrisRackauckas:
Okay cool, a lot of it worked. Can you fix up that last bit? Then we should do a quick merge so the majority of it is in.

PR #223, opened 2023-09-19
Add support of `missing` values for `L2Loss`

amostof:


ChrisRackauckas:
Can you add a test?

src/cost_functions.jl
```diff
@@ -50,14 +50,14 @@ function (f::L2Loss)(sol::DiffEqBase.AbstractNoTimeSolution)
 
     if weight == nothing
         @inbounds for i in 1:length(sol)
-            sumsq += (data[i] - sol[i])^2
+            sumsq += (coalesce(data[i] - sol[i], 0))^2
```
Vaibhavdixit02, line 53: use zero of the same type as elements

PR #290, opened 2026-04-13
Support SciMLBase v3

ChrisRackauckas-Claude:
## Summary

- Replace `DiffEqBase.DEProblem` with `SciMLBase.AbstractDEProblem` in `multiple_shooting_objective` and `two_stage_objective`. The bare `DEProblem` alias was removed in SciMLBase v3.
- Port the ensemble tutorial to the new v3 `prob_func(prob, ctx)` signature, using `ctx.sim_id` instead of the `i` argument.
- Extend `SciMLBase` compat to `"1.69, 2, 3.1"` and bump `DiffEqParamEstim` to `2.4.0`.

Supersedes the Project.toml-only bump in #289.

## Known limitation — please read before merging

**Local `Pkg.resolve()` fails** on any environment that contains SciMLBase 3.1, because `SciMLBase` 3.1 pins `RecursiveArrayTools = 4` while `OrdinaryDiffEq` in the registry still pins `RecursiveArrayTools ≤ 3.54`. CI on this PR will reproduce the same unsatisfiable resolution until upstream `OrdinaryDiffEq` releases a RAT-v4-compatible version.

That means:

- I **have not** been able to run tests locally. Per our usual rule of not claiming something works without running it, this PR should be treated as unverified until CI passes.
- CI is expected to fail at the `Pkg.add` / `instantiate` step with an `Unsatisfiable requirements detected for package RecursiveArrayTools` error, **not** at Julia runtime. Once `OrdinaryDiffEq` bumps, re-running CI here should succeed without further edits.

## What is NOT in this PR (and why)

The `sol[i]` / `length(sol)` / `for x in sol` call sites in `src/cost_functions.jl` and `src/multiple_shooting_objective.jl` are **deliberately untouched**. I could not conclusively verify in the SciMLBase v3.1 source whether scalar integer indexing on `AbstractTimeseriesSolution` changed behavior — no explicit `Base.getindex(::AbstractTimeseriesSolution, ::Int)` override is present there. Rewriting the cost-function machinery on an unverified claim would be reckless.

Once CI on this PR actually exercises those code paths (either passing silently or producing concrete failures), the real scope of any follow-up work will be visible, and we can land a second PR for cost-function indexing if needed.

## Test plan

- [ ] CI passes once OrdinaryDiffEq releases a RAT-v4-compatible version.
- [ ] Specifically, verify the ensemble tutorial build succeeds with the new `(prob, ctx)` signature.
- [ ] Triage any runtime failures from cost-function indexing before merging; file follow-up PR if needed.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DiffEqParamEstim"
 uuid = "1130ab10-4a5a-5621-a13d-e4788d82bd4c"
-version = "2.3.0"
+version = "2.4.0"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
 
 [deps]
@@ -27,7 +27,7 @@ ForwardDiff = "0.10"
 PenaltyFunctions = "0.1, 0.2, 0.3"
 PreallocationTools = "0.2, 0.3, 0.4, 1.0"
 RecursiveArrayTools = "1.0, 2.0, 3, 4"
-SciMLBase = "1.69, 2"
+SciMLBase = "1.69, 2, 3.1"
 SciMLSensitivity = "7"
 Statistics = "1.10"
 StatsAPI = "1.8.0"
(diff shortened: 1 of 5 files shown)


ChrisRackauckas-Claude:
cc @ChrisRackauckas

PR #295, opened 2026-04-27
Fix BlackBoxOptim test calls to use Optimization.jl problem interface

ChrisRackauckas-Claude:
## Summary

`build_loss_objective` and `multiple_shooting_objective` now return `OptimizationFunction`, but several tests still called `bboptimize` directly with lowercase `search_range`/`max_steps` kwargs. BlackBoxOptim does not normalize those into its internal `:SearchRange`/`:MaxSteps` keys, so it fell back to the default 1-D `(-1.0, 1.0)` `SearchRange` and threw `ArgumentError: You MUST specify NumDimensions=` from `check_and_create_search_space`. This has been the cause of the long-standing CI failure across `lts`, `1`, and `pre` Julia versions following the migration to `OptimizationFunction` (PR #293).

This PR converts the affected tests (`test/tests_on_odes/l2loss_test.jl`, `test/tests_on_odes/blackboxoptim_test.jl`, `test/multiple_shooting_objective_test.jl`) to build an `Optimization.OptimizationProblem` with `lb`/`ub` derived from the existing tuple bounds and solve it via `OptimizationBBO`'s `BBO_adaptive_de_rand_1_bin_radiuslimited()` — the same pattern already used in `test/likelihood.jl`. Test assertions are updated to use `result.u` instead of the BlackBoxOptim-specific `archive_output.best_candidate`.

## Reproduction

```julia
julia> using BlackBoxOptim
julia> bboptimize(x -> sum(abs2, x .- 1.5); search_range = [(1.0, 2.0)], max_steps = 100)
ERROR: ArgumentError: You MUST specify NumDimensions= in a solution when giving a SearchRange=(-1.0, 1.0)
```

(`search_range`/`max_steps` are silently ignored — only the `:SearchRange`/`:MaxSteps` keys are honored, but the canonical SciML way is to go through `OptimizationProblem`.)

## Test plan

- [x] `runic --check src test` passes on the modified files.
- [x] Verified the new `OptimizationProblem`/`solve(..., BBO_adaptive_de_rand_1_bin_radiuslimited())` pattern works in isolation against an `OptimizationFunction`.
- [ ] CI green on `lts`, `1`, and `pre`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -41,6 +41,7 @@ BlackBoxOptim = "a134a8b2-14d6-55f6-9291-3336d3ab0209"
 DelayDiffEq = "bcd4f6db-9728-5f36-b5f7-82caef46ccdb"
 ExplicitImports = "7d51a73a-1435-4ff3-83d9-f097790105c7"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 NLopt = "76087f3c-5699-56af-9a33-bf431cd00edd"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 Optimization = "7f7a1694-90dd-40f0-9382-eb1efda571ba"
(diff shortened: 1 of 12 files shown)


ChrisRackauckas:
Heads-up from an automated CI green-up pass: the `runic` check is failing with `runic-action: julia is a required dependency but does not seem to be available`. The `fredrikekre/runic-action@v1` step does not install Julia itself, so `.github/workflows/FormatCheck.yml` needs a `setup-julia` step before it:

```diff
--- a/.github/workflows/FormatCheck.yml
+++ b/.github/workflows/FormatCheck.yml
@@ -14,6 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: '1'
       - uses: fredrikekre/runic-action@v1
         with:
           version: '1'
```

I could not push this to the PR head branch directly (this environment is restricted to `SciML/*`, not the `ChrisRackauckas-Claude` fork that hosts the PR head).

Separately, the `test (alldeps, 1.10)` Downgrade job is failing during `julia-actions/julia-buildpkg@v1` (compat-resolution failure with the lower bounds), which looks pre-existing and orthogonal to the BlackBoxOptim fix in this PR.

The `lts` / `1` / `pre` jobs are currently in-progress on `09e02ed`; I'll keep monitoring.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #308, opened 2026-06-13
Use SciMLTesting v1.2 (folder-based run_tests)

ChrisRackauckas-Claude:
Convert the test suite to the SciMLTesting v1.2 folder-discovery model. `test/runtests.jl` becomes:

```julia
using SciMLTesting
run_tests()
```

## Layout
- **Core** = top-level `test/*.jl`. The former `test/tests_on_odes/*.jl` files are flattened to the top level and made self-contained — each now loads `DiffEqParamEstim` + its own deps and `include`s the shared problem fixture — so each runs isolated in its own `@safetestset` module (folder-discovery requires self-contained files; they previously shared one `@testset` scope).
- **test/shared/** (a non-group folder, never auto-discovered) holds the shared `test_problems.jl` fixture, the never-referenced `lorenz_test.jl` / `lorenz_true_test.jl`, and the commented-out `genetic_algorithm_test.jl`. This keeps the discovered Core set byte-for-byte equal to the prior dispatch.
- **QA** = `test/qa/` with its existing sub-env `Project.toml`.

## Deps
- Add `SciMLTesting` + `SafeTestsets` to the root test target and the QA sub-env `Project.toml`.
- Drop `Pkg` from the root test deps (only the old harness used it; no test file references it).
- `test/test_groups.toml` unchanged.

## Verification
- Static: every `Project.toml` + `test_groups.toml` parses; every test `.jl` parses; folder-discovery (real SciMLTesting v1.2.0) resolves Core/All to exactly the 16 files the old dispatcher ran and QA to `qa.jl` in its sub-env.
- End-to-end: `GROUP=QA` `Pkg.test()` on Julia 1.11 passed (7 pass / 2 broken — the 2 pre-existing `@test_broken`), exercising the sub-env activation path.
- Runic-formatted.

Ignore until reviewed by @ChrisRackauckas.


ChrisRackauckas-Claude:
Force-updated this branch to the corrected **SciMLTesting v1.1.0** conversion, superseding the v1.0.0 first pass.

What changed vs the v1.0.0 commit on this branch:
- **QA fixed.** v1.0.0 put `"QA"` in `groups=`, but `"QA"` is a reserved key `run_tests` intercepts before the groups table — with no `qa=` body, `GROUP=QA` threw `ArgumentError`. Now QA uses the reserved `qa=` kwarg and is selectable by name.
- **Core body is now a file path** (`test/core_tests.jl`) instead of an inline `() -> begin ... end` thunk (the v1.1.0 world-age-safe, isolated `@safetestset` form).
- **Explicit `all = ["Core"]`** documents and pins the `All` membership (Core only, QA omitted) rather than relying on the legacy fallback.

Behavior is identical to upstream `master` for every `GROUP` value (bare/All, Core, QA). Static-verified: `TOML.parsefile` on both `Project.toml` files and `Meta.parseall` on `runtests.jl`/`core_tests.jl`/`qa/qa.jl` all clean.

PR #309, opened 2026-06-14
Canonicalize test units to @safetestset

ChrisRackauckas-Claude:
Converts the independent top-level test units in `test/runtests.jl` from plain `@testset "X" begin ... end` to `@safetestset`, so each runs in its own module (test isolation + world-age safety), matching OrdinaryDiffEq's canonical structure.

**Converted units:** `Explicit Imports`, `Tests on ODEs`, `Multiple Shooting Objective`, `Likelihood Loss`, `Out-of-place ODE Tests`, `Steady State Tests`, `DAE Tests`, `DDE Tests`, `Test on Monte`.

**Self-containment handling:**
- `Tests on ODEs` is a single shared-state unit: `tests_on_odes/test_problems.jl` defines `prob1`/`prob2`/`prob3`/`t`/`data`/`sol` that the later includes in the same block consume, and several of those includes rely on `Optim`/etc. leaking between files within the block. It is therefore wrapped as **one** `@safetestset` whose body is the include chain (preserving the intra-block sharing exactly). Added `using DiffEqParamEstim` at the top of that body since the included files relied on it leaking from `Main`; SafeTestsets injects `using Test`.
- `out_of_place_odes.jl` used `build_loss_objective`/`L2Loss` but only had `using OrdinaryDiffEq, Test, ...` and relied on `DiffEqParamEstim` leaking from `Main` → added `DiffEqParamEstim` to its `using` line.
- The other standalone files already import `DiffEqParamEstim` plus their optimizer/solver deps, so they are self-contained as `@safetestset` modules.

**Deps:** added `SafeTestsets` to `[extras]`, `[targets].test`, and `[compat]` (`SafeTestsets = "0.1, 1"`), and `using SafeTestsets` in `runtests.jl`.

The `GROUP`/`QA` dispatch and all assertions are unchanged.

Verification: static — `Meta.parseall` of `runtests.jl` and `out_of_place_odes.jl` parse cleanly, and Runic formatting passes on all changed files. The suite pulls a heavy optimization/sensitivity stack (BlackBoxOptim, NLopt, Optim, Optimization, SciMLSensitivity, Sundials, StochasticDiffEq, Zygote) with long-running global optimizers, so it was not run locally; CI will verify.

This PR should be ignored until reviewed by @ChrisRackauckas.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -39,6 +39,7 @@ PenaltyFunctions = "0.3"
 PreallocationTools = "0.4.34, 1.0"
 Random = "1"
 RecursiveArrayTools = "3.50, 4"
+SafeTestsets = "0.1, 1"
 SciMLBase = "2.153, 3"
 SciMLSensitivity = "7.80"
 SteadyStateDiffEq = "2"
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
Superseded by the v1.2 folder conversion on sciml-testing-rollout (#308).

